PR #12, opened 2017-06-15
Generic pread and API enhancement

m4b:
Experimental API refactor + generics branch

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,8 +59,7 @@ let byte: scroll::Result<i64> = b.pread(0);
 let slice = b.pread_slice::<str>(0, 2).unwrap();
 let byte_slice: &[u8] = b.pread_slice(0, 2).unwrap();
 
-// here is an example of parsing a uleb128 custom datatype, which
-// uses the `ctx::DefaultCtx`
+// here is an example of parsing a uleb128 custom datatype
 let leb128_bytes: [u8; 5] = [0xde | 128, 0xad | 128, 0xbe | 128, 0xef | 128, 0x1];
 // parses a uleb128 (variable length encoded integer) from the above bytes
 let uleb128: u64 = leb128_bytes.pread::<scroll::Uleb128>(0).unwrap().into();
--- a/README.md
+++ b/README.md
@@ -73,7 +73,7 @@ assert_eq!("hello_world", hello_world);
 
 // ... and this parses the string if its space separated!
 let spaces: &[u8] = b"hello world some junk";
-let world: &str = spaces.pread_with(6, ctx::SPACE).unwrap();
+let world: &str = spaces.pread_with(6, StrCtx::Delimiter(SPACE)).unwrap();
 assert_eq!("world", world);
 ```
 
--- a/examples/data_ctx.rs
+++ b/examples/data_ctx.rs
@@ -8,12 +8,12 @@ struct Data<'a> {
     id: u32,
 }
 
-impl<'a> ctx::TryFromCtx<'a, (usize, Endian)> for Data<'a> {
+impl<'a> ctx::TryFromCtx<'a, Endian> for Data<'a> {
     type Error = scroll::Error;
-    fn try_from_ctx (src: &'a [u8], (offset, endian): (usize, Endian))
+    fn try_from_ctx (src: &'a [u8], endian: Endian)
                      -> Result<Self, Self::Error> {
-        let name = src.pread::<&'a str>(offset)?;
-        let id = src.pread_with(offset+name.len()+1, endian)?;
+        let name = src.pread::<&'a str>(0)?;
+        let id = src.pread_with(name.len()+1, endian)?;
         Ok(Data { name: name, id: id })
     }
 }
--- a/src/ctx.rs
+++ b/src/ctx.rs
@@ -67,7 +67,7 @@ use core::str;
 use core::result;
 
 use error;
-use endian;
+use endian::Endian;
 
 pub trait MeasureWith<Ctx> {
     type Units;
@@ -90,12 +90,6 @@ impl<Ctx, T: AsRef<[u8]>> MeasureWith<Ctx> for T {
         self.as_ref().len()
     }
 }
-
-/// The default parsing context; use this when the context isn't important for your datatype
-pub type DefaultCtx = endian::Endian;
-
-/// Convenience constant for the default parsing context
-pub const CTX: DefaultCtx = endian::NATIVE;
 
 /// The parsing context for converting a byte sequence to a `&str`
 ///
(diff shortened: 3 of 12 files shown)


m4b:
@goandylok

I added you to collaborators so you should be able to push directly to this branch. I think first thing would be to add your Str enum back into the branch, thus giving back slice functionality. Then I can refactor to tests to use this and get them working again. I'll probably start on converting the Pwrite side and related into Traits to use new generic form with offsetting logic in pwrite. 

Note on commit messages, I prefer to have the module/whatever be preceded by message when the commit affects it. This makes it very easy to scan git log for changes what happened and what it affected. 

If there are multiple ones I use a comma. If there are only small changes in different modules I usually leave it out, or if the effect is related to the changes to the module I leave just the main module affected.

goandylok:
Oho! It seems nice to use `Index<RangeFrom<I>>` for the new offset logic. And there are also changes of `Error` -- adding `TooBig` variant and generic variables. But there are some points I am not sure: 
- What's the purpose of `From` bound for `Error`? Are there any requirement that wants to cast custom type to `scroll::Error`?
- Are the error variants `BadRange` and `BadInput` reasonable to exist?
- What is the behavior of `TooBig { size: T, len: T }` when returned by a struct consisting from small primitives? Should the size be the same of the incomplete primitive's or the same of the struct's?
- What role does `MeasureWith` have?
- Which experiment features should we adopt now?
  - ~~New `pread` offset~~ 
  - Custom generic error variant
  - Returning size
  - Removing `FromCtx`
  - Grouping `Pread` with `Gread`
  - `TryFromCtx` for bool
  - Only implement `Pread` for AsRef<[u8]> and not for &'a [u8]

m4b:
Sorry should have mentioned that bare minimum was done to get compiling without error and at least run tests (maybe not pass)

Basically errors will all go away except a few. Really busy right now will respond in depth later, but keep in mind if something looks weird I probably hacked it so the crate would compile 🤣

m4b:
> What's the purpose of From bound for Error? Are there any requirement that wants to cast custom type to scroll::Error?

This was the generic error issue i mentioned above; it's one way to solve it, but requires users to impl `From` scroll for their own errors (which they would always do anyway I believe)

> Are the error variants BadRange and BadInput reasonable to exist?

no probably not; only there to allow compilation, we can evaluate which are reasonable later once more parts are integrated.

> What is the behavior of TooBig { size: T, len: T } when returned by a struct consisting of small primitives? Should the size be the same of the incomplete primitive's or the same of the struct's?

This is a good question :) but i believe that is up to the implementor; or they can use a custom error and return whatever they want.

> What role does MeasureWith have?

It's the only way I could get completely generic pread working (need to know how big `Self` is)

m4b:
1. [x] New pread offset
2. [x] Custom generic error variant - I think this is done?
3. [ ] Returning size - i believe this is the future, but want more testing/integration see how it behaves and performs
4. [ ] Removing FromCtx - i don't want to do this unless we can (I don't know if we can, we shall see) 
5. [ ] Grouping Pread with Gread (and Cread and Lread) - yes please!
6. [ ] TryFromCtx for bool - i don't see why not, but semantics aren't clear?
7. [ ] Only implement Pread for AsRef<[u8]> and not for &'a [u8] - i believe it's implemented for everything that implements those bounds now with super generic version, so this shouldn't be as much a worry anymore

Here's some other things I want to see/think will be possible:

0. `gread` version for `cread`
1. if we can get a single `use scroll::Pread` which includes `cread`, `pread`, `gread`,  and `lread`, i think that would be so amazing on so many levels
2. single derive! wow this would be so awesome, so that `#[derive(Pread)]` gets you `gread, lread, cread, and pread` (because single trait)
3. I think I want to explore `pread` slices of types too. this comes up frequently for me. in particular, i want to write `let x: &'a [u64] = bytes.pread::<&[u64]>(0, 10);` which would more or less be equivalent to a `let x: &'a [u64] = ::std::mem::transmute(bytes.as_ptr() as *const u64, 10)`; what's fun is that we can fallback to default context for numbers (e.g., machine endianness) which afaiu is the only correct way to transmute a byte slice to a slice of numbers

goandylok:
> TryFromCtx for bool - i don't see why not, but semantics aren't clear?

In experiment repo, the semantic of reading bool now is `pread::<u8>()? != 0`, and writing is `pwrite::<u8>(u8::max_value())`

> Custom generic error variant - I think this is done?

No, there are not generic on custom variant so far. The new `Error` you drafted only allow users to **read** custom error, but not to **raise** custom error. Assume that users want to a parsing data with special  standard to struct, which could be invalid even if all it's primitives are `pread`ed successfully, users can't return a error type specify to this data format. What I purpose here is to generic over `Error::Custom(E)`.

> This is a good question :) but i believe that is up to the implementor; or they can use a custom error and return whatever they want.

IMO, we should remove variables of `TooBig`.

> I think I want to explore pread slices of types too. 

Here is some tricky things come in. I was actually working on this before; I gave scroll ability to read and write slices of any `T: TryFromCtx<Ctx>`; Reading such a slice requires ctx like `(SliceCtx, Ctx)`, where `Ctx` is used by inner `T` and `SliceCtx` is a enum containing three variants: `Eager`, `Length(usize)` and `EagerUntil(usize)`, I think it's easy to guess its semantic by name.

By the way, I want to add the third variant for `StrCtx`: `DelimiterUntil(usize)` for the case of reading &str that has fixed length but may be delimited early.

goandylok:
Forget to say that `(SliceCtx, Ctx)` solution have may drawbacks:

- Pread returns a iterator instead of slice
```rust
bytes.pread::<scroll::ctx::slice::Iter<&str>>(0, (SliceCtx::Eager, StrCtx::Delimiter(0)));
```
- It check the length twice.

So I removed this part at last.

m4b:
so `Pwrite` is totally generic now, and all tuple offset logic is removed :tada: 

> By the way, I want to add the third variant for StrCtx: DelimiterUntil(usize) for the case of reading &str that has fixed length but may be delimited early.

This is a good idea; comes up in unix archives and PE binaries quite frequently actually, as the string is usually 16 bytes but space or null padded

> No, there are no generic on custom variant so far. The new Error you drafted only allow users to read custom error, but not to raise custom error. Assume that users want to a parsing data with special standard into struct, which could be invalid even if all it's primitives are preaded successfully, users can't return a error type specify to this data format. So what I purpose here is to generic over Error::Custom(E)

So I tried this originally, but found custom error while defining `TryFromCtx` sufficient for my needs; https://github.com//m4b/goblin/blob/c42dd67989490ace89e513eb99eff350d60f63dc/src/elf/header.rs#L306 is an example of custom error defined by client and returned

m4b:
RE slices, I feel like it should be possible, but maybe I'm missing something; have to play with it.

m4b:
So next step I think you should add the StrCtx enum style, update `TryFromCtx` with the impl, add tests for new methods of reading, etc.

Also after than can swap out all uses of `pread_slice::<str>` with `pread_with::<&str>`

After this, can start investigating the size based returns and merging Gread into Pread

goandylok:
Ok, I will take over these stuffs.

m4b:
@goandylok ok! G, L, and C are all back and all tests are passing/compiling (minus gread_inout, etc., and scroll_derive no longer working)

So funny thing, and the error message is terrible, but when you `cursor.lread::<Type>()`, the context passed to `SizeWith` _must_ match the context passed to the `FromCtx` impl, otherwise it cannot statically dispatch correctly (and gives compile time error); I believe this is more indication that your approach of returning the size in the `TryFromCtx/FromCtx` impl is the correct approach semantically, as the above then cannot happen.

Again, more experimentation needed, but I think this branch is ready to see that.

m4b:
Another super cool feature of making everything generic is that tryoffset with is gone now.

m4b:
@goandylyk I might merge this, any objections? (I'll cleanup Error first, remove unused variants if haven't already)

Can you open another PR for size returning variants of TryFromCtx? I believe it will be possible

goandylok:
Sure, I will. It's better to fix the `Error` stuffs before we merge.

m4b:
So; i don't see the point of adding another generic type parameter to `Error`; it would only be used by us, since we're the ones who would use it to embed other errors in `TryFromCtx` impls; otherwise, if a user wants to use a custom error, they are free to simply use their own custom error type using the associated type; unless I'm missing something.

I do think maybe some or all of the parameters need to be removed in errors; need to test actual output; in my experience, having the offset the error occurred at was _extremely_ useful, and I'm actually pretty sad it's gone now...

I'm going to merge this though

PR #13, opened 2017-06-16
New StrCtx

goandylok:
This branch is __not__ ready to merge. Some roadblock is obstructing this refactor.

### DIFF
--- a/src/ctx.rs
+++ b/src/ctx.rs
@@ -101,29 +101,25 @@ pub const CTX: DefaultCtx = endian::NATIVE;
 ///
 /// `StrCtx` specifies what byte delimiter to use, and defaults to C-style null terminators. Be careful.
 #[derive(Debug, Copy, Clone)]
-pub struct StrCtx {
-    pub delimiter: u8
+pub enum StrCtx {
+    Delimiter(u8),
+    DelimiterUntil(u8, usize),
+    Length(usize),
 }
 
 /// A C-style, null terminator based delimiter for a `StrCtx`
-pub const NULL: StrCtx = StrCtx { delimiter: 0 };
+pub const NULL: StrCtx = StrCtx::Delimiter(0);
 /// A space-based delimiter for a `StrCtx`
-pub const SPACE: StrCtx = StrCtx { delimiter: 0x20 };
+pub const SPACE: StrCtx = StrCtx::Delimiter(0x20);
 /// A newline-based delimiter for a `StrCtx`
-pub const RET: StrCtx = StrCtx { delimiter: 0x0a };
+pub const RET: StrCtx = StrCtx::Delimiter(0x0a);
 /// A tab-based delimiter for a `StrCtx`
-pub const TAB: StrCtx = StrCtx { delimiter: 0x09 };
+pub const TAB: StrCtx = StrCtx::Delimiter(0x09);
 
 impl Default for StrCtx {
     #[inline]
     fn default() -> Self {
         NULL
-    }
-}
-
-impl From<u8> for StrCtx {
-    fn from(delimiter: u8) -> Self {
-        StrCtx { delimiter: delimiter }
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/ctx.rs
```diff
@@ -121,12 +123,6 @@ impl Default for StrCtx {
     }
 }
 
-impl From<u8> for StrCtx {
-    fn from(delimiter: u8) -> Self {
-        StrCtx { delimiter: delimiter }
-    }
-}
-
```
goandylok: Removed because semantic is vague to casting u8 between StrCtx

src/ctx.rs
```diff
@@ -349,39 +345,29 @@ macro_rules! into_ctx_float_impl {
...
+            }
+            StrCtx::Length(len) => len,
+        };
+
+        if len > src.len() {
+            return Err(error::Error::TooBig{size: len, len: src.len()});
+        };
+
+        str::from_utf8(&src[..len])
+            .map_err(|_| error::Error::Custom("utf8 error".to_owned()))
```
goandylok: `str::from_utf8` returns `Utf8Error` but I can't express it due to poor generic support for `Error::Custom` variance

src/ctx.rs
```diff
@@ -121,12 +123,6 @@ impl Default for StrCtx {
     }
 }
 
-impl From<u8> for StrCtx {
```
goandylok: Removed because semantic is vague to casting `u8` between `StrCtx`

src/ctx.rs
goandylok: `str::from_utf8` returns `Utf8Error` but I can't express it due to poor generic support for `Error::Custom` variance

goandylok:
As `Null` delimited str is not a default in rust, and we can't say any one of the variants is accepted generically in most use case, we should remove `Default` for `StrCtx` in order to force user to provide one.

src/ctx.rs
m4b: Why? only variant this could be is `delimiter`

src/ctx.rs
m4b: In original scroll I believe I had bad input with static string; we can't use custom here because it allocates and would force this entire impl to be behind a feature flag (also this won't compile now in `no_std`); I understand now what you mean by `From` being a bound, since it means we can't use std lib or core errors.

So we can either add an extra variant like original, or try out your generic version.

m4b:
> As Null delimited str is not a default in rust, and we can't say any one of the variants is accepted generically in most use case, we should remove Default for StrCtx in order to force user to provide one.

I understand this point of view but I don't think I'm willing to make this change. From a selfish viewpoint, it would break too much of my code, and I don't feel like dealing with it.

Secondly, whenever I parse a str from a sequence of bytes, it is almost always in my experience a null-terminated string.

I think it is a pretty reasonable default for parsing out a str from a sequence of bytes, and as long as it's documented I don't think it's unreasonable.

And lastly, it's just extraordinarily convenient, nice to use, which is the ultimate goal of scroll.

However, down the line I could get behind implementing `CStr` for scroll, and switching over the expected defaults to this and maybe removing (or deprecating somehow) them for `&str` (as well as automatic + 1 to len writes).

src/ctx.rs
goandylok: Think about what will happen: 
```rust
let str: &str = bytes.pread_with(20.into());
```

src/ctx.rs
goandylok: I mean not `From` bound, but generic_error (https://github.com/goandylok/scroll2/pull/1/commits/39ed1a3677e97d35198e5db21535fc9d3a1ef378)

```rust
pub enum Error<E> {
   BadOffset(usize),
   Incomplete,
   Other(E),
   // IO(std::io::Error)
}
```

src/ctx.rs
m4b: Length takes a usize, delim a u8. But I see your point will be ambiguous to a reader perhaps or will bite you if not careful. On that note I thought we would allow str reading via length by simply passing usize? I think this way is really cool and reasonable

goandylok: The type annotation may be even longer than a enum variance. :-/ It's really tricky to distinguish behaviour by integer type. If you really want to read fixed-length str by simply usize, you can impl `TryFromCtx<'a, &[u8], usize>` for str, which can coexist with the StrCtx one.

m4b: Ya the latter is what I meant. It's not a big deal can add it later.

PR #14, opened 2017-06-16
Remove default ctx

goandylok:


goandylok:
I think you should not rely on default ctx type like `Endian`, as what you said in document.
> `DefaultCtx`	The default parsing context; use this when the context isn't important for your datatype

And I would like to remove bound `Ctx: Default` of `pread()`. Actually, we can't `pread()` without providing type annotation, which means that we are forced to write the following code:
```rust
bytes.pread::<bool, DefaultCtx>(0);
```
To fix it, we could force `pread()` only to be used on type with ctx `()`.

m4b:
I don't understand; current version doesn't require that at all, it works as usual with default, no?

```
let astr = bytes.pread::<&str>(0)?;
let int = bytes.pread::<u64>(0)?;
```

should work, and all tests pass and compile.

Removing `Default` to allow contextless parsing is a non-starter for me; 2f54064 moves it to the bound on `pread` to make it less strict but default parsing is so important for ergonomics that any changes which endanger this simply aren't worth it to me.

m4b:
> I think you should not rely on default ctx type like Endian, as what you said in document.

So the point of `FromCtx<Ctx = Endian>` is just pure convenience on side of implementor; you still have to type `from_ctx(&[u8], ctx: Endian)` so it won't lead to weird bugs, or unknown ctxs; it just makes the burden of implementor one less line, since almost every time I use `FromCtx` or `TryFromCtx` I always want the `Endian` ctx to pass into `pread_with`.

On the other hand though, if the user wants to just read at machine endianness, then they will likely just use `pread::<u64>` in the `TryFromCtx` body, which means that setting the parsing context to `TryFromCtx<()>` is reasonable; so perhaps it's fine to not have a default type parameter in `FromCtx` and `TryFromCtx`.

goandylok:
> I don't understand; current version doesn't require that at all, it works as usual with default, no?

Sorry, I don't realize this code could work. But this is strange to me, how compiler knows the type? It fails in my experiment repo. And this test also falls after I add the second impl for &str with unit ctx.

m4b:
> But this is strange to me, how compiler knows the type?

Through careful magiks, I arranged this situation :laughing: 

The trick is you can't have two Ctx impls for the same type, iirc; so this probably makes parsing out u64 with `usize` ctx not feasiable; now that i think about it, I think this is why i added the original `pread_slice` method, to allow parsing slices

m4b:
P.S. almost done re-adding `Gread/Gwrite` back into repo; after refactoring the code, I believe that your `TryFromCtx` returning the size will work, but must see how it interacts with everything; as I said, `scroll` is a cruel mistress

goandylok:
> Through careful magiks, I arranged this situation 

Awesome and impressive. Is it black magic in rust? 😱

PR #15, opened 2017-06-17
Fix bug for StrCtx

goandylok:
`StrCtx::DelimiterUntil` stopped at the end of slice when no delimiter found nor length satisfied.


goandylok:
Another issue:
Delimiter consts can't be used by `DelimiterUntil`.

Fix:
Make the consts u8 but the drawback is that it forces `StrCtx::Delimiter(Null)`.

m4b:
For simple changes like this please push directly to generic_pread branch

PR #26, opened 2018-02-08
Run the examples in the README as doctests

luser:
I don't know if you actually want this, but I wanted to write an example of using scroll_derive, and I realized that the examples in the README weren't being tested. I wired them up with my current best skeptic approximation that doesn't pull in any dependencies, but it requires some fiddling with the examples because they don't get an automatic `extern crate scroll` for some reason when run this way. They wind up looking slightly uglier, but it does ensure that they work!

The second commit then adds an example with derive, and tweaks your Travis CI config to build with `--features=derive` so it works. One quirky thing here is that if someone does `cargo test` without that feature, that test will fail to compile. I don't know if that concerns you or not. If it does, I could make that work by splitting it out to a separate markdown file.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,3 +6,7 @@ rust:
 matrix:
   allow_failures:
     - rust: nightly
+
+script:
+  - cargo build --verbose --features=derive
+  - RUST_BACKTRACE=1 cargo test --all --verbose --features=derive
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
  [![Build Status](https://travis-ci.org/m4b/scroll.svg?branch=master)](https://travis-ci.org/m4b/scroll)
 ## Scroll - cast some magic
 
-```
+```text
          _______________
     ()==(              (@==()
          '______________'|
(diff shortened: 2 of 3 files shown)


m4b:
First off: thank you, this is totally awesome !

Want to merge ASAP just need quick clarification about:

> One quirky thing here is that if someone does cargo test without that feature, that test will fail to compile

It sounds like not a big deal but not sure exactly what you mean. 

Also would making the derive feature in the default features help this ?  I’m considering switching it to default and making people who don’t need it use no default; not sure. 

Need to decide for next release about feature soon, because I’d like the new documentation here showing up :)

luser:
> It sounds like not a big deal but not sure exactly what you mean.

If you simply do `cargo test` with these changes, it'll fail. If you do `cargo test --features=derive` it'll pass. The problem is that given the hacky way I'm running the tests, I don't have any way to make the doctests in README.md conditional. Putting `derive` in the default features would fix that so that it'd only break if you did `cargo test --no-default-features`, which seems like a pretty unlikely thing for people do to.

Alternately, as I mentioned, if we moved the derive tests to a separate markdown file, we could have a separate test that's marked `#[cfg(feature = "derive")]` and it'd work properly, but then it wouldn't be in the README which would be a bummer.

I guess I could also file a bug on rustdoc asking for a way to have cfg directives for doctests.

PR #28, opened 2018-05-02
tests: fix safe_packed_borrows warning

### LINKED ISSUE #24: fix tests due to packed issue
`assert_eq` borrows the packed structs which is now Against The Rules

ref https://github.com/rust-lang/rust/issues/46043

hdhoang:
Close #24


m4b:
@hdhoang Hey thanks for your PR, this is great!

Would not have thought to use scopes for a quick fix, great work, thanks again!

PR #30, opened 2018-08-30
Serialize/Deserialize a std::ffi::CStr

Michael-F-Bryan:
This PR adds implementations of `TryIntoCtx` and `TryFromCtx` for `&CStr` and `CString` when compiling with the `std` feature enabled. 

We could probably speed the `TryFromCtx` up a little bit by using something like `memchr` for finding the string's null terminator seeing as that's the most expensive part, but that adds an extra dependency and I wasn't sure whether it was necessary.

### DIFF
--- a/src/ctx.rs
+++ b/src/ctx.rs
@@ -548,6 +548,40 @@ impl<'a> TryFromCtx<'a> for CString {
     }
 }
 
+#[cfg(feature = "std")]
+impl<'a> TryIntoCtx for &'a CStr {
+    type Error = error::Error;
+    type Size = usize;
+    #[inline]
+    fn try_into_ctx(self, dst: &mut [u8], _ctx: ()) -> error::Result<Self::Size> {
+        let data = self.to_bytes_with_nul();
+
+        if dst.len() < data.len() {
+            Err(error::Error::TooBig {
+                size: dst.len(),
+                len: data.len(),
+            })
+        } else {
+            unsafe {
+                copy_nonoverlapping(data.as_ptr(), dst.as_mut_ptr(), data.len());
+            }
+
+            Ok(data.len())
+        }
+    }
+}
+
+#[cfg(feature = "std")]
+impl TryIntoCtx for CString {
+    type Error = error::Error;
+    type Size = usize;
+    #[inline]
+    fn try_into_ctx(self, dst: &mut [u8], _ctx: ()) -> error::Result<Self::Size> {
+        self.as_c_str().try_into_ctx(dst, _ctx)
+    }
+}
+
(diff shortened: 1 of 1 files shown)


src/ctx.rs
```diff
@@ -513,6 +516,72 @@ impl TryIntoCtx<Endian> for usize where usize: IntoCtx<Endian> {
...
+impl<'a> TryIntoCtx for &'a CStr {
+    type Error = error::Error;
+    type Size = usize;
+    #[inline]
+    fn try_into_ctx(self, dst: &mut [u8], _ctx: ()) -> error::Result<Self::Size> {
+        let data = self.to_bytes_with_nul();
+
+        if dst.len() < data.len() {
+            Err(error::Error::TooBig {
+                size: dst.len(),
```
m4b: These field names are confusing. I should fix this before 1.0. I can’t even remember which is which

src/ctx.rs
```diff
@@ -513,6 +516,72 @@ impl TryIntoCtx<Endian> for usize where usize: IntoCtx<Endian> {
...
+    fn try_from_ctx(src: &'a [u8], _ctx: ()) -> result::Result<(Self, Self::Size), Self::Error> {
+        let null_byte = match src.iter().position(|b| *b == 0) {
+            Some(ix) => ix,
+            None => return Err(error::Error::BadInput {
+                size: 0,
+                msg: "The input doesn't contain a null byte",
+            })
+        };
+
+        let data = src.get(..null_byte+1).unwrap();
```
m4b: Why not just `src[..null_byte+1]`?

Also so I’m too stupid right now to reason about the +1 indexes and since there are tests checking bytes read I’ll just assume it’s correct :P

Michael-F-Bryan: I remember hitting a really bizarre lifetimes issue, where `&src[..null_byte + 1]` would reborrow from `src` giving `data` a temporary lifetime which only lasted until the end of the function. The `get()` method returned an `&[u8]` with the `'a` lifetime we want to return.

I believe I was writing this PR using `nightly` and when I tried to reproduce it in the playground everything worked :disappointed: Maybe there was a bug in NLL that's been fixed recently?

Also, the `+1` is because `..` is a half open range. When constructing the `&CStr` via `CStr::from_bytes_with_nul_unchecked(data)` we want to include that trailing `null`, so you need to specify the index *after* our null terminator.

src/ctx.rs
Michael-F-Bryan: I was a little confused too. Would it be better to call them something like `expected_length` and `actual_length`?

PR #32, opened 2018-09-07
Add support for u128/i128. Fixes #31.

### LINKED ISSUE #31: Add support for u128/i128
`u128`/`i128` are missing from the list of primitives that get implementations of all the nice traits in scroll.

luser:
This change adds u128 and i128 to the list of primitive types that get
implementations of all the basic scroll traits.


m4b:
@luser awesome thanks for this! Sure glad I wrote those macros now :D

PR #35, opened 2018-09-13
Implement IntoCtx/TryIntoCtx for references to primitive types.

luser:
This should help in fixing https://github.com/m4b/scroll_derive/issues/15 .

### DIFF
--- a/src/ctx.rs
+++ b/src/ctx.rs
@@ -188,6 +188,12 @@ macro_rules! into_ctx_impl {
                 write_into!($typ, $size, self, dst, le);
             }
         }
+        impl<'a> IntoCtx<Endian> for &'a $typ {
+            #[inline]
+            fn into_ctx(self, dst: &mut [u8], le: Endian) {
+                (*self).into_ctx(dst, le)
+            }
+        }
         impl TryIntoCtx<Endian> for $typ where $typ: IntoCtx<Endian> {
             type Error = error::Error;
             type Size = usize;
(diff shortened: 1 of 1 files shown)


src/ctx.rs
```diff
@@ -188,6 +188,12 @@ macro_rules! into_ctx_impl {
                 write_into!($typ, $size, self, dst, le);
             }
         }
+        impl<'a> IntoCtx<Endian> for &'a $typ {
+            #[inline]
+            fn into_ctx(self, dst: &mut [u8], le: Endian) {
+                (*self).into_ctx(dst, le)
```
m4b: Ah i see; so this is sort of a "stopgap" solution, and still won't work for non-copy types.

I was also thinking the other day it would be cool to have a pread_ref and it safely performs a mem::transmute for you and ties the reference to the bytes; this came up recently when attempting to pread a huge object onto the stack and blowing up my stack; what i needed was a reference (and it didn't copy), but had to resort to mem::transmute, and check length, and bla bla; seems like scroll could safely automate some of that drudgery

luser: I looked into something like this a while back before deciding that scroll was the sensible option--I thought that what I really wanted was to be able to just transmute references to structs to avoid copying entirely. The recent changes to make references to fields in packed structs unsafe convinced me otherwise, though. You'd have to ensure that the alignment was correct and for structs they'd have to be packed or contain no padding and it just all seems like a hassle.

luser: For the record:

> Ah i see; so this is sort of a "stopgap" solution, and still won't work for non-copy types.

This patch only covers the primitive numeric types in scroll, `scroll_derive` already generates impls for `&'a T` as well as `T`, and the latter calls the former so it should be fine for non-`Copy` types:
https://github.com/m4b/scroll_derive/blob/d8fe9d0a5a7ea21823ce9fd6206766ae81782e4a/src/lib.rs#L95

Additionally https://github.com/m4b/scroll_derive/pull/17 fixes those impls to always use references so that non-`Copy` struct fields will be handled properly.

PR #38, opened 2018-09-20
Mention that the derive feature needs to be enabled in the README

luser:
I've stumbled over this basically every time I've used scroll. :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,6 +116,13 @@ fn main() {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


m4b:
Yea sorry I made the feature optional because I don’t want to force non std users to use it, though I don’t know if this is a concern in practice ?

_Or_ if you only want the Pread pwrite implementations (eg for reading numbers out, you don’t add a bunch of compile time through syn, etc). Maybe it deserves some attention to determine if it should be automatic (a la the great pioneer crate, failure). Dunno

PR #43, opened 2018-10-01
fix some clippy warnings

llogiq:


src/ctx.rs
```diff
@@ -109,24 +108,26 @@ impl Default for StrCtx {
...
     }
+
+    pub fn is_empty(&self) -> bool {
+        if let StrCtx::Length(_) = *self { true } else { false }
+    }
 }
 
 /// Reads `Self` from `This` using the context `Ctx`; must _not_ fail
 pub trait FromCtx<Ctx: Copy = (), This: ?Sized = [u8]> {
-    #[inline]
```
m4b: What is the rationale for removing inline annotation here? Does it have no effect when it's on the trait def? (I've always wondered actually).

Otherwise, we def should have inline, as perf goes down pretty significantly without it

llogiq: Inline annotations are only effective on impls.

llogiq:
You're welcome.

Issue #37, opened 2018-09-20
Reasoning behind associated `Size` / `Units` types on various traits?


luser:
I've stumbled over this before while writing functions that were generic over `TryIntoCtx` and `SizeWith`--the constraints need to explicitly call out `TryIntoCtx<Size=usize>` or `SizeWith<Units=usize>` in order to be able to actually *use* the sizes for anything meaningful.

Both `MeasureWith` and `SizeWith` have an associated `Units` type:
https://docs.rs/scroll/0.9.0/scroll/ctx/trait.MeasureWith.html#associatedtype.Units
https://docs.rs/scroll/0.9.0/scroll/ctx/trait.SizeWith.html#associatedtype.Units

And both `TryFromCtx` and `TryIntoCtx` have an associated `Size` type:
https://docs.rs/scroll/0.9.0/scroll/ctx/trait.TryFromCtx.html#associatedtype.Size
https://docs.rs/scroll/0.9.0/scroll/ctx/trait.TryIntoCtx.html#associatedtype.Size

I'm sure you had an intended use case for this being generic when you wrote these, but is it actually something that has proven useful in practice? grepping through the `scroll`, `scroll_derive` and `goblin` repositories shows that there aren't any implementations of these traits using anything but `usize`, and most usage in `goblin` is via the derives, which all provide implementations where the types are `usize`.

I know it would be a breaking change, but removing these associated types and just using `usize` everywhere seems like it would simplify things a bit.

The only counter-argument I can come up with on my own is that perhaps you might want to have types that have sizes that don't fit in 32 bits but you want to use them with these traits in 32-bit code, but this feels a bit contrived and also would suggest that just using `u64` instead of `usize` everywhere would be the answer.

m4b:
Will write thoughts more on weekend but briefly:

1. I’m open to removing it
2. It will probably be a lot of work

Original usecase of generic index was that you might want to use some other mechanism to index into something other than bytes (think Index trait); however in practice implementing this in scroll itself never materialized (non generic bytes itself were used as the final indexee in the various Try* traits).

As for SizeWith I can’t remember why I added that associated type ? I probably had generic fever ? Was worried about something ? Added it because I couldn’t get the Pread traits to compile ?

Is MeasureWith used anywhere ? That might be accidentally included from old prototypes or maybe it’s used.

[...]

PR #45, opened 2018-10-18
Remove generic {Size,Units} in favor of usize

willglynn:
This PR addresses https://github.com/m4b/scroll/issues/37, globally squishing `{Size,Units}` into `usize`. This is a breaking change which warrants careful consideration, but… tests pass.

Corresponding `scroll_derive` PR: https://github.com/m4b/scroll_derive/pull/19

### DIFF
--- a/README.md
+++ b/README.md
@@ -194,10 +194,9 @@ struct Data<'a> {
 // note the lifetime specified here
 impl<'a> ctx::TryFromCtx<'a, Endian> for Data<'a> {
   type Error = scroll::Error;
-  type Size = usize;
   // and the lifetime annotation on `&'a [u8]` here
   fn try_from_ctx (src: &'a [u8], endian: Endian)
-    -> Result<(Self, Self::Size), Self::Error> {
+    -> Result<(Self, usize), Self::Error> {
     let offset = &mut 0;
     let name = src.gread::<&str>(offset)?;
     let id = src.gread_with(offset, endian)?;
--- a/examples/data_ctx.rs
+++ b/examples/data_ctx.rs
@@ -10,9 +10,8 @@ struct Data<'a> {
 
 impl<'a> ctx::TryFromCtx<'a, Endian> for Data<'a> {
     type Error = scroll::Error;
-    type Size = usize;
     fn try_from_ctx (src: &'a [u8], endian: Endian)
-                     -> Result<(Self, Self::Size), Self::Error> {
+                     -> Result<(Self, usize), Self::Error> {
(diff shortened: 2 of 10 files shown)


luser:
This looks like what I expected, so that's good. :)

willglynn:
The only thing I can think of is this sacrifices the ability to scroll through 64-bit indexes on platforms where `usize` is 32 bits. This isn't as serious as it sounds since anyone using `#[derive(Pread)]` is scrolling through `&[u8]`s which are indexed by `usize` anyway.

In principle someone could be doing 64-bit offset math and then transforming that into `mmap()`ed `&[u8]`s pointing to different sections of a file or something _while also_ using their own handwritten `TryFromCtx`/`FromCtx` implementations, but that seems far fetched.

philipc:
In gimli, we have a [`Reader::Offset`](https://github.com/gimli-rs/gimli/blob/5398258cd2dd8fdf8ef092ce612d5cc9919ac171/src/reader.rs#L203) whose purpose is to allow 32-bit platforms to read 64-bit debuginfo over the network (though I'm not aware of anyone doing this yet). I don't know if scroll intends to be used for that kind of case, but as @willglynn says, it would need to be reading from something other than `&[u8]`, which doesn't seem to have good support in scroll.

m4b:
I want to merge this but why is the CI failing?

willglynn:
This PR depends on the `scroll_derive` PR to be green, since the README tests `#[derive(Pread, Pwrite)]`. Likewise, the `scroll_derive` PR depends on this PR for the type changes.

Tests passed when I had both checked out locally and overrode `{ path = "../…" }` in `Cargo.toml`, but there's no easy way to get Travis to do test both branches against each other.

m4b:
Hmm ok. This will need a new 8.0 release. I think should probably be after the moving to non-owned pwrites. 

(Btw this recently came up for me; I needed to box an extremely large object, but I couldn’t see an easy or immediate way to pwrite the object without blowing up my stack.)

willglynn:
That other change is another case where `scroll`+`scroll_derive` need to have changes merged simultaneously. Maybe `scroll_derive` should live in `scroll`'s repo?

m4b:
mokay i moved scroll_derive into scroll, and nuked the git history in the process, but, um i guess that's fine? looking at failure + failure_derive it looks like a similar thing happened, and they didn't update/remove the failure_derive repo. we probably shouldn't do that here? e.g., should point back here? i guess? i dunno.

PR #47, opened 2018-11-02
Make [Try]IntoCtx borrow instead of copying

### LINKED ISSUE #42: Packed structs and Scroll
E.g., this is already going to be a problem: https://github.com/m4b/goblin/issues/105

Need to figure out path forward for packed structs (and if there is any work that scroll needs to do to ameliorate the situation)

### LINKED ISSUE #46: IntoCtx/TryIntoCtx on borrowed values
The trait signatures are:

```rust
/// Writes `Self` into `This` using the context `Ctx`
pub trait IntoCtx<Ctx: Copy = (), This: ?Sized = [u8]>: Sized {
    fn into_ctx(self, &mut This, ctx: Ctx);
}

/// Tries to write `Self` into `This` using the context `Ctx`
pub trait TryIntoCtx<Ctx: Copy = (), This: ?Sized = [u8]>: Sized {
    type Error;
    fn try_into_ctx(self, &mut This, ctx: Ctx) -> Result<usize, Self::Error>;
}
```

These traits mirror `Into` which consumes `self`. However, these might be the wrong function signatures for how `scroll` is used in practice; both traits are primarily used on borrowed values.

`scroll` [implements `IntoCtx`](https://docs.rs/scroll/0.9.2/scroll/ctx/trait.IntoCtx.html#foreign-impls) and [`TryIntoCtx`](https://docs.rs/scroll/0.9.2/scroll/ctx/trait.IntoCtx.html#foreign-impls) on borrowed primitive types, suggesting that `[Try]IntoCtx` needn't consume its value after all.

`scroll_derive` generates code [implementing `IntoCtx`](https://github.com/m4b/scroll_derive/blob/master/src/lib.rs#L307-L312) and [`TryIntoCtx`](https://github.com/m4b/scroll_derive/blob/master/src/lib.rs#L111-L113) on `Self` by borrowing and delegating to the implementation on `&'a Self`.

Should these traits take `&self` instead of `self`?

willglynn:
This PR changes `[Try]IntoCtx` to take `&self` instead of `self`. It very likely breaks every downstream project that calls `cwrite()` or `pwrite()`, but in a trivial way – callers need only specify `&`.

This closes #46 and also incidentally closes #42.

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,7 +163,7 @@ fn write_io() -> Result<(), scroll::Error> {
     let mut bytes = [0x0u8; 10];
     let mut cursor = Cursor::new(&mut bytes[..]);
     cursor.write_all(b"hello")?;
-    cursor.iowrite_with(0xdeadbeef as u32, BE)?;
+    cursor.iowrite_with(&(0xdeadbeef as u32), BE)?;
     assert_eq!(cursor.into_inner(), [0x68, 0x65, 0x6c, 0x6c, 0x6f, 0xde, 0xad, 0xbe, 0xef, 0x0]);
     Ok(())
 }
--- a/scroll_derive/examples/main.rs
+++ b/scroll_derive/examples/main.rs
@@ -18,7 +18,7 @@ fn main () {
     println!("data: {:?}", &data);
     assert_eq!(data.id, 0xdeadbeefu32);
     let mut bytes2 = vec![0; ::std::mem::size_of::<Data>()];
-    bytes2.pwrite_with(data, 0, LE).unwrap();
+    bytes2.pwrite_with(&data, 0, LE).unwrap();
     let data: Data = bytes.pread_with(0, LE).unwrap();
     let data2: Data = bytes2.pread_with(0, LE).unwrap();
     assert_eq!(data, data2);
--- a/scroll_derive/src/lib.rs
+++ b/scroll_derive/src/lib.rs
@@ -91,22 +91,14 @@ fn impl_try_into_ctx(name: &syn::Ident, fields: &syn::FieldsNamed) -> proc_macro
     }).collect();
     
     quote! {
-        impl<'a> ::scroll::ctx::TryIntoCtx<::scroll::Endian> for &'a #name {
+        impl ::scroll::ctx::TryIntoCtx<::scroll::Endian> for #name {
             type Error = ::scroll::Error;
             #[inline]
-            fn try_into_ctx(self, dst: &mut [u8], ctx: ::scroll::Endian) -> ::scroll::export::result::Result<usize, Self::Error> {
+            fn try_into_ctx(&self, dst: &mut [u8], ctx: ::scroll::Endian) -> ::scroll::export::result::Result<usize, Self::Error> {
                 use ::scroll::Pwrite;
                 let offset = &mut 0;
                 #(#items;)*;
                 Ok(*offset)
-            }
-        }
-
-        impl ::scroll::ctx::TryIntoCtx<::scroll::Endian> for #name {
-            type Error = ::scroll::Error;
-            #[inline]
-            fn try_into_ctx(self, dst: &mut [u8], ctx: ::scroll::Endian) -> ::scroll::export::result::Result<usize, Self::Error> {
-                (&self).try_into_ctx(dst, ctx)
             }
         }
     }
@@ -275,14 +267,14 @@ fn impl_into_ctx(name: &syn::Ident, fields: &syn::FieldsNamed) -> proc_macro2::T
                 quote! {
                     let size = ::scroll::export::mem::size_of::<#arrty>();
                     for i in 0..self.#ident.len() {
-                        dst.cwrite_with(self.#ident[i], *offset, ctx);
+                        dst.cwrite_with(&self.#ident[i], *offset, ctx);
                         *offset += size;
                     }
                 }
             },
             _ => {
                 quote! {
-                    dst.cwrite_with(self.#ident, *offset, ctx);
+                    dst.cwrite_with(&self.#ident, *offset, ctx);
                     *offset += #size;
                 }
             }
@@ -290,20 +282,13 @@ fn impl_into_ctx(name: &syn::Ident, fields: &syn::FieldsNamed) -> proc_macro2::T
     }).collect();
 
     quote! {
-        impl<'a> ::scroll::ctx::IntoCtx<::scroll::Endian> for &'a #name {
+        impl ::scroll::ctx::IntoCtx<::scroll::Endian> for #name {
             #[inline]
-            fn into_ctx(self, dst: &mut [u8], ctx: ::scroll::Endian) {
+            fn into_ctx(&self, dst: &mut [u8], ctx: ::scroll::Endian) {
                 use ::scroll::Cwrite;
                 let offset = &mut 0;
                 #(#items;)*;
                 ()
-            }
-        }
-
-        impl ::scroll::ctx::IntoCtx<::scroll::Endian> for #name {
-            #[inline]
-            fn into_ctx(self, dst: &mut [u8], ctx: ::scroll::Endian) {
-                (&self).into_ctx(dst, ctx)
             }
         }
     }
--- a/scroll_derive/src/lib.rs
+++ b/scroll_derive/src/lib.rs
@@ -8,6 +8,38 @@ extern crate syn;
 
 use proc_macro::TokenStream;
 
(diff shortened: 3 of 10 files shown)


willglynn:
I'm working up a big PR for `goblin` that addresses this and the earlier `Size`/`Units` change, which seems like a representative way to evaluate the impact. (Also, I broke it/I bought it?)

willglynn:
I opened PRs in both `goblin` and `faerie`. Good thing too; I had missed updating the `iowrite()` trait.

README.md
```diff
@@ -163,7 +163,7 @@ fn write_io() -> Result<(), scroll::Error> {
     let mut bytes = [0x0u8; 10];
     let mut cursor = Cursor::new(&mut bytes[..]);
     cursor.write_all(b"hello")?;
-    cursor.iowrite_with(0xdeadbeef as u32, BE)?;
+    cursor.iowrite_with(&(0xdeadbeef as u32), BE)?;
```
m4b: This and the 3 others are a huge PR, will take me a bit to review, appreciate your patience, @willglynn.
I would like to note that _having_ to take the reference now of variables, literals, makes me pretty sad...
I feel like it jars slightly with the original and basic purpose of scroll, which was to "serialize" numbers and PODs.
I do wish there had been a little more discussion before this enormous (and amazing PR!), specifically, whether we can have our cake and eat it too w.r.t. TryInto for the primitives.
Specifically, I'm wondering if there's a compromise, so that we can still pass either references _or_ owned values to the pwrite/iowrite.
Maybe it's just me but i do find it a usability paper cut and find it less attractive that I have to take references to pwrite now, but perhaps I'm used to the old way.
So, is there some generic magic we can use to do this? Guessing wildy and randomly:
1. Have someone take AsRef
2. Have scroll_derive derive the reference versions for any user type (i'm ok with taking references of user structs)
3. ???

On the other hand, in my case where the struct was too large to be put on the stack, and I had to box it, even if we did say 2., this would still probably eventually blow the stack? Or perhaps not. Have to think about it, and its late and a Thursday :)
Similarly, and against my original sadness, if I'm able to pwrite out a huge boxed struct that auto-derives pwrite, that would actually make me happy and would counteract any bitterness of no longer being able to pwrite raw deadbeefs :D
Dunno, just some thoughts, feel free to respond whenever you get the time (and that goes for anyone else who's watch).
Regardless, thanks for your amazing work here @willglynn and the robustness and tenacity of the changes, great work!

willglynn: Cheers! And yeah, this is why I went and did these PRs even with limited discussion: I didn't know how this concept would look until it was done 😄

I'll ponder the generics question. There might be a way we can make `pwrite()` be `&`-agnostic while keeping `TryIntoCtx` take `&self` – that'd let you `pwrite()` either by value or by reference without consuming `self` during the write.

willglynn:
If we say:

```rust
    /// Write `n` of type `T` at offset `I` with context `Ctx`
    /// # Example
    /// ```
    /// use scroll::{Pwrite, Pread, LE};
    /// let mut bytes: [u8; 8] = [0, 0, 0, 0, 0, 0, 0, 0];
    /// bytes.pwrite_with(0xbeefbeef, 0, LE).unwrap();
    /// assert_eq!(bytes.pread_with::<u32>(0, LE).unwrap(), 0xbeefbeef);
    fn pwrite_with<T,V>(&mut self, n: V, offset: usize, ctx: Ctx) -> result::Result<usize, E>
        where T: TryIntoCtx<Ctx, <Self as Index<RangeFrom<usize>>>::Output, Error = E> + ?Sized,
              V: Borrow<T>,
    {
        let len = self.measure_with(&ctx);
        if offset >= len {
            return Err(error::Error::BadOffset(offset).into())
        }
        let dst = &mut self[offset..];
        n.borrow().try_into_ctx(dst, ctx)
    }
```

These work:

* `bytes.pwrite_with(0xbeefbeef, 0, LE)`
* `bytes.pwrite_with(&0xbeefbeef, 0, LE)`

This does not work, because now there's two type parameters:

* `bytes.pwrite_with::<u32>(0xbeefbeef, 0, LE)`

These also work:

* `bytes.pwrite_with::<u32,_>(0xbeefbeef, 0, LE)`
* `bytes.pwrite_with::<_,u32>(0xbeefbeef, 0, LE)`
* `bytes.pwrite_with::<u32,u32>(0xbeefbeef, 0, LE)`

philipc:
Using `Borrow` makes sense to me. Does `pwrite_with::<...>` get used much in practice? I guess it's only for constants, and for those I think `bytes.pwrite_with(0xbeefbeef_u32, 0, LE)` is clearer.

m4b:
I'm pretty excited about this version @willglynn ! I checked out your branch this weekend and played around with it, if nothing else comes up, I think this might be the way forward :)

So we can now do stuff like this:

```rust
#[test]
fn pwrite_big_struct() {
    use scroll::{LE, Pwrite};
    #[derive(Pread, Pwrite, Copy, Clone)]
    struct Big {
        arr: [u8; 100000000],
    }
    let big: Box<Big> = box (unsafe { ::std::mem::uninitialized() });//Big { arr: [0; 100000000] };
    let mut bytes = vec![0; 100000001];
    //let big = box bytes.pread_with::<Big>(0, LE).unwrap();
    let _ = bytes.pwrite_with(&* big, 0, LE).unwrap();
    assert!(false);
}
```

which was my original pain point when trying to write a large struct which was boxed :D (Preading is still an issue, and boxing doesn't fix it atm, but I think we could in principle fix that too, at another time)

Just grepping through a couple of codebases, I believe @philipc is correct (as usual) and afaics only goblin atm uses `pwrite::<T>` and this is because it relies on eliding an `into()` call, which is easily fixed (and probably better with an explicit into anyway):

```
src/mach/symbols.rs
277:                (bytes.pwrite_with::<Nlist32>(self.into(), 0, le)?)
280:                (bytes.pwrite_with::<Nlist64>(self.into(), 0, le)?)

src/mach/segment.rs
201:            bytes.pwrite_with::<Section64>(self.into(), 0, ctx.le)?;
203:            bytes.pwrite_with::<Section32>(self.into(), 0, ctx.le)?;
374:            bytes.pwrite_with::<SegmentCommand64>(self.into(), 0, ctx.le)?;
376:            bytes.pwrite_with::<SegmentCommand32>(self.into(), 0, ctx.le)?;
```

So yea, overall really like this direction, seems like with a minor sacrifice we can keep the owned usage for literals (and other Copyables?). If we do end up going this route, I think we'll need some explicit documentation about why passing a type parameter via turbofish is banned/problematic, etc.

Anyway, great work !

willglynn:
Cool! I plan to:

- [ ] Keep the `[Try]IntoCtx` traits taking `&self`,
- [ ] Keep the `scroll_derive` changes around `&self`,
- [ ] Change the `Pwrite` and `IOwrite` traits to use `Borrow` as [described above](
https://github.com/m4b/scroll/pull/47#issuecomment-435549006),
- [ ] Reset everything that's not in the list above,
- [ ] Fix everything that's broken

I'll comment when that's done and I've force-pushed the results.

m4b:
Friendly ping @willglynn . Would definitely like to see this get pushed through? Do you have a branch WIP anywhere, or need to offload some work to someone else? Hoping we didn't lose steam on it as i think its definitely something that's needed :)

m4b:
No worries, thanks for the update! :)

m4b:
@willglynn gentle ping :) is there any chance of continuing here, or should we explore other options? If you're busy don't worry at all, or don't want to work on it anymore, don't feel bad at all/responsible whatever, just trying to get a feel for how to proceed/what to do.
If you get a chance let me know; maybe someone else can pick it up or I can take a look at it, etc.
As I understood it its basically working, modulo using borrow to deal with the ergonomic issue of pwriting constants?

willglynn:
Sorry! I've been traveling some more and haven't got back to this. Probably the right thing to do is to reset to `master` to integrate the latest changes, change the interfaces [as described upthread](https://github.com/m4b/scroll/pull/47#issuecomment-436872219), and then just chase the build errors until they're gone again. This is still a thing I'm interested in doing, but if someone else beats me to it, that'd be cool too :+1:

m4b:
gosh i don't even know where to begin to move this as a PR into the scroll repo :(

m4b:
I cargo fmt scroll and scroll derive; was hoping something like this would get resolved, but doesn't seem like it's going to happen, and now that fmt has happened there will be conflicts everywhere. going to close this; we can revisit again, or possibly never :)

Issue #40, opened 2018-09-23
Scroll 0.9.1 doesn't compile under Rust 1.25.0 due to i128


sunfishcode:
Cranelift currently supports compiling with Rust 1.25.0, and it uses Faerie which uses Scroll. The Scroll 0.9.1 update doesn't compile under Rust 1.25.0, since it uses i128, added in https://github.com/m4b/scroll/pull/32, which isn't supported in that version of Rust.

<details>
<summary>Sample compilation error</summary>

```
error[E0658]: 128-bit type is unstable (see issue #35118)
   --> /home/me/.cargo/registry/src/github.com-1ecc6299db9ec823/scroll-0.9.1/src/ctx.rs:223:38
    |
223 |         impl<'a> FromCtx<Endian> for $typ {
    |                                      ^^^^

```

</details>

<br>
I can work around this for now by just fixing the dependency to scroll 0.9.0, and I'll look into other options on Cranelift side, but I'm interested if you have any ideas here too.

philipc:
build.rs could automatically define a feature for the rust version and use that to enable i128 support, similar to [what the rand crate does](https://github.com/rust-random/rand/blob/9a4e62ab4463b88ba642aef3d72977aecb4e2ef0/src/distributions/integer.rs#L84).

m4b:
goddamnit; should i yank and publish as 0.10?

m4b:
I broke goblin too so that's great also :/

sunfishcode:
Would you be interested in a PR implementing @philipc's idea? That would require adding a build.rs file, but it can follow [rand's version](https://github.com/rust-random/rand/blob/master/build.rs) which is pretty simple.

m4b:
Maybe; I'd almost prefer to publish a 0.10.0 and let everyone wait until they upgrade/the new edition comes out, and then say it requires 2018; but anyway, i also don't see how that build.rs would fix things? I imagine it's more involved?

m4b:
not sure if yanking and publishing 0.10.0 would work though; also no sure how i ever published a newer version because scroll has a circular dependency on itself, heh.

m4b:
Ok, published 0.9.2; cranelift-faerie is compiling again, thanks to everyone helping out here!

m4b:
Scroll requires 1.20 now; we'll remove the build script once rust 2018 edition is out, and require rust 2018

PR #49, opened 2019-04-05
Upgrade to 2018 edition

lzutao:
This PR is based on decision from #40 of @m4b : https://github.com/m4b/scroll/issues/40#issuecomment-425742488


m4b:
@lzutao thank you for this! I want to land this but I really wanted to see @willglynn PR here landed first, which was a really great improvement, namely it allows:
1. pwriting extremely large values out of boxes, instead of on the stack (which will usually blow up the stack)
2. better user/api ergonomics by not requiring owned values in pwrite calls.

Let me see what status is, if not maybe I or someone else can try to rebase and fix.

m4b:
actually this probably has minor conflicts? let me check

PR #68, opened 2020-04-13
Add generics support to scroll_derive

Systemcluster:
This adds support for deriving `Pread`, `Pwrite`, `IOread`, `IOwrite` and `SizeWith` on generic structs.

### DIFF
--- a/scroll_derive/tests/tests.rs
+++ b/scroll_derive/tests/tests.rs
@@ -158,6 +158,28 @@ fn test_nested_struct() {
     assert_eq!(b, b2);
 }
 
+#[derive(Debug, PartialEq, Eq, Pread, Pwrite, IOread, IOwrite, SizeWith)] 
+#[repr(C)]
+struct Data8<T, Y> {
+    ids: [T; 3],
+    xyz: Y
+}
+
+#[test]
+fn test_generics() {
+    let mut bytes = [0xde, 0xed, 0xef, 0x10, 0x10];
+    let data: Data8<u8, u16> = bytes.pread_with(0, LE).unwrap();
+    assert_eq!(data.ids, [0xde, 0xed, 0xef]);
+    assert_eq!(data.xyz, 0x1010);
+    let data: Data8<u8, u16> = bytes.cread_with(0, LE);
+    assert_eq!(data.ids, [0xde, 0xed, 0xef]);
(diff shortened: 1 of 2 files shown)


m4b:
Shit was I supposed to merge other first :(

m4b:
Ok looks like you just have to rebase @Systemcluster according to comments in other; thank you for this btw, this is quite cool!

Systemcluster:
Rebased, thanks for merging these!

scroll_derive/tests/tests.rs
```diff
@@ -158,6 +158,28 @@ fn test_nested_struct() {
...
+#[repr(C)]
+struct Data8<T, Y> {
+    ids: [T; 3],
+    xyz: Y
+}
+
+#[test]
+fn test_generics() {
+    let mut bytes = [0xde, 0xed, 0xef, 0x10, 0x10];
+    let data: Data8<u8, u16> = bytes.pread_with(0, LE).unwrap();
```
m4b, line 171: what happens if we do:
```rust
let data: Data8<Data8<u8, 16>, u16> = bytes.pread_with()
```
is this just a flat packed serialize/deserialize ?

Systemcluster: That won't work because of the trait bounds for the generic params in the impl, but I think it would be useful if it were possible. I'll see if I can figure out a way to allow for that.

m4b: Yea i'm just wondering what would happen; and also my refrain that this isn't meant to be a general purpose serialization/deserialization library, but i suppose that cat might have gotten out already ;)

Systemcluster: You'd just get an error like `the trait bound Data8<u8, 16>: std::convert::From<u8> is not satisfied` with this implementation.

m4b:
@Systemcluster do you want a new release? I think this is safe for a patch release, since it just adds additional deriving functionality, and shouldn't be able to break anyone?

m4b:
(Also thanks very much for taking the time to add this functionality, very cool stuff :))

m4b:
ok scroll derive 0.10.2 is out, so should be able to use this from crates.io now; thanks for the hard work! :)

PR #73, opened 2020-08-17
Fix links and outdated information in doc comments

dequbed:
This removes the reference to the `Gread` trait (rolled into `Pread`)
and updates the links to the `FromCtx`, `IntoCtx` and `TryFromCtx`
traits.


m4b:
This is awesome, thank you !

PR #74, opened 2020-08-18
Restructuring the documentation

### LINKED ISSUE #60: What does the "g" in "gread" stand for?
Fun fact: the docs for the `gread` method do not even contain the letter "g" :)

>Reads a value from self at offset with a default Ctx. For the primitive numeric values, this will read at the machine's endianness. Updates the offset

I was able to figure out the meaning of `cread` and `pread`, but `gread` escapes me. What does the "g" stand for?

### LINKED ISSUE #67: Docs mention lost Gread and Gwrite traits
It appears that the trait `Gread` was demoted to oblivion in  605455704d95aebb36f114bcd79d7533609775e9; however, various parts of the API documentation still mention it. Here are two examples:

https://github.com/m4b/scroll/blob/e94a84d374161ccd3b966e0684fb72de6471e8a8/src/lib.rs#L17-L21
https://github.com/m4b/scroll/blob/8a205d8e22465b494c1cf7a02d8fe1034f044df2/src/ctx.rs#L140

dequbed:
I've recently started working with scroll and found the documentation somewhat hard to follow.

Since complaining about cool stuff I get for free (which this library definitely qualifies for) isn't all that nice I'm trying to improve the docs and structure them to be easier to follow for would-be users.

Please don't merge this PR yet; since I only just started using the library I may be wrong about some details of how stuff works; I'm mostly opening this PR to allow for easier comments and criticism on what I'm writing.

This PR will also fix #60 and fix #67 when done.

### DIFF
--- a/src/ctx.rs
+++ b/src/ctx.rs
@@ -258,6 +258,78 @@ pub trait FromCtx<Ctx: Copy = (), This: ?Sized = [u8]> {
 }
 
 /// Tries to read `Self` from `This` using the context `Ctx`
+///
+/// # Implementing Your Own Reader
+/// If you want to implement your own reader for a type `Foo` from some kind of buffer (say
+/// `[u8]`), then you need to implement
+///
+/// ```rust
+/// use scroll::{self, ctx, Pread};
+/// #[derive(Debug, PartialEq, Eq)]
+/// pub struct Foo(u16);
+///
+/// impl<'a> ctx::TryFromCtx<'a, scroll::Endian> for Foo {
+///      type Error = scroll::Error;
+///      fn try_from_ctx(this: &'a [u8], le: scroll::Endian) -> Result<(Self, usize), Self::Error> {
+///          if this.len() < 2 { return Err((scroll::Error::Custom("whatever".to_string())).into()) }
+///          let n = this.pread_with(0, le)?;
+///          Ok((Foo(n), 2))
+///      }
+/// }
+///
+/// let bytes: [u8; 4] = [0xde, 0xad, 0, 0];
+/// let foo = bytes.pread_with::<Foo>(0, scroll::LE).unwrap();
+/// assert_eq!(Foo(0xadde), foo);
+///
+/// let foo2 = bytes.pread_with::<Foo>(0, scroll::BE).unwrap();
+/// assert_eq!(Foo(0xdeadu16), foo2);
+/// ```
+///
+/// # Advanced: Using Your Own Error in `TryFromCtx`
+/// ```rust
+///  use scroll::{self, ctx, Pread};
+///  use std::error;
+///  use std::fmt::{self, Display};
+///  // make some kind of normal error which also can transform a scroll error ideally (quick_error, error_chain allow this automatically nowadays)
+///  #[derive(Debug)]
+///  pub struct ExternalError {}
+///
+///  impl Display for ExternalError {
+///      fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
+///          write!(fmt, "ExternalError")
+///      }
+///  }
+///
+///  impl error::Error for ExternalError {
+///      fn description(&self) -> &str {
+///          "ExternalError"
+///      }
+///      fn cause(&self) -> Option<&error::Error> { None}
+///  }
+///
+///  impl From<scroll::Error> for ExternalError {
+///      fn from(err: scroll::Error) -> Self {
+///          match err {
+///              _ => ExternalError{},
+///          }
+///      }
+///  }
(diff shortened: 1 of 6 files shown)


src/lib.rs
```diff
@@ -12,127 +12,199 @@
...
+//!
+//! # Examples
+//!
+//! Let's start with a simple example
 //!
 //! ```rust
-//! use scroll::{ctx, Pread, LE};
+//! use scroll::{ctx, Pread};
+//!
+//! // Let's first define some data, cfg-gated so our assertions later on hold.
```
dequbed, line 57: Notably, I have not actually checked if these examples work on big endian architectures. ^^'

src/lib.rs
```diff
@@ -12,127 +12,199 @@
...
+//! #[cfg(target_endian = "little")]
 //! let bytes: [u8; 4] = [0xde, 0xad, 0xbe, 0xef];
+//! #[cfg(target_endian = "big")]
+//! let bytes: [u8; 4] = [0xef, 0xbe, 0xad, 0xde];
```
dequbed, line 61: Notably, I have not actually checked if this example works on big endian architectures. ^^'

m4b:
This looks great already, thanks for doing something like this! What are your plans/needs to move this from WIP -> reviewable/mergeable state?

dequbed:
Sorry, I was rather slammed with ... life mostly the last days

> This looks great already, thanks for doing something like this! What are your plans/needs to move this from WIP -> reviewable/mergeable state?

I'd say it's about 50% done. I want to in the end have touched up just about every file to make sure nothing is outdated and in the best case added a few examples although I think I'll do that in a seperate PR to get this one done sooner. If everything goes according to plan I'll have this one done by the end of the week ^^'

dequbed:
I feel with the last commit that this is at a state where I'm happy with it to be included. The examples are not all updated yet but I would address that in a separate PR otherwise this one will not get finished for a while longer.

src/lib.rs
```diff
@@ -12,127 +12,201 @@
...
-//! assert_eq!("hello_world", hello_world);
+//! assert_eq!("hello world", hello_world);
 //!
-//! // ... and this parses the string if its space separated!
-//! use scroll::ctx::*;
-//! let spaces: &[u8] = b"hello world some junk";
-//! let world: &str = spaces.pread_with(6, StrCtx::Delimiter(SPACE)).unwrap();
-//! assert_eq!("world", world);
+//! // We can again provide a custom context; for example to parse Space-delimited strings.
+//! let hello2: &[u8] = b"hello world\0more words";
```
m4b, line 100: hmm, this example confused me at first, but i see now it illustrates both space delimited parsing, and the null terminator is correctly seen as single item, so it's good

m4b (changes requested):
sorry a lot of comments; most of them are nits, but the typos would be nice to fix. if you don't have time, please feel free to let me know, and we can merge as is, and I can clean it up. If you are busy, please don't hesitate to tell me, it's ok! you already did a lot of great work here and should feel happy and i want to merge this :) Thank you! :heart:

src/ctx.rs
```diff
@@ -121,6 +258,78 @@ pub trait FromCtx<Ctx: Copy = (), This: ?Sized = [u8]> {
...
+/// let foo2 = bytes.pread_with::<Foo>(0, scroll::BE).unwrap();
+/// assert_eq!(Foo(0xdeadu16), foo2);
+/// ```
+///
+/// # Advanced: Using Your Own Error in `TryFromCtx`
+/// ```rust
+///  use scroll::{self, ctx, Pread};
+///  use std::error;
+///  use std::fmt::{self, Display};
+///  // make some kind of normal error which also can transform a scroll error ideally (quick_error, error_chain allow this automatically nowadays)
```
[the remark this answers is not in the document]
dequbed: You're right, I totally missed that comment in a comment ^^'

dequbed:
Thanks for spellchecking me a few of those I should totally have catched.
I made one commit for each file but if you prefer those two - or the entire PR - squashed I can do that too.

src/lib.rs
dequbed: Yeah reading it a few days after writing it it's not so obvious what I'm  trying to demonstrate - I hope the expanded comment helps with clarity.

m4b:
I just read through the rendered version, it is very nice, thank you so much for your patience and putting some love into the docs, incredibly appreciated :heart:

I'm going to squash all the changes if you don't mind, it's a nice holistic PR this way

m4b (approved):
LGTM, thanks again for this, seriously :)

Issue #56, opened 2019-09-06
remove usize pread/pwrite


m4b:
this was a terrible mistake and should never have been done (by me).

m4b:
this is now removed as of dfd9c7b730d0661688d2b86c90400ee0dfb1b3c9 and will be included in 0.11

Issue #51, opened 2019-04-06
Design: boxed preads


m4b:
In anticipation of #47, the final remaining piece for working with arbitrary data safely is to figure out how to pread large objects without blowing the stack.
Most system languages accomplish this by boxing the object and manipulating it through the heap pointer.
I'm wondering if we can accomplish this in a semi-automated way, without compromising scroll's interesting type based reading/writing.

Doing this properly would allow us to round trip a large stack object within scroll.

Specifically, could we do something like:
```rust
#[derive(Pread, Pwrite, Copy, Clone)]
struct Big {
    arr: [u8; 100000000],
}
let mut bytes = vec![42; 100000001];
let big = box bytes.pread_with::<Box<Big>>(0, LE)?; // this is boxed
// by the scroll Box impl, using unsafe unitialized memory
// which is only returned if the pread succeeds, so it remains safe?
let _ = bytes.pwrite_with(big, 0, LE)?; // Note: there is no `&*`,
// ideally scroll can deal with this case, though I don't know if its possible?
```

In worst case, we should be able to expose a new `pread_with_box` method, that does the above, but pushes the box type information into the method name. This is fine, but i think its a more beautiful api design to allow the user to choose which type to read out via turbofish.

cc @philipc @lzutao @willglynn @luser 
What you all think?

philipc:
Do you have a real world example that would use this?

m4b:
Yes, similar to the example above, I was dealing with some C structs which had some header information, etc., then a large byte array containing image bytes, then some more stuff.
I wanted to pread the struct, match on some header values, validate, then do something with the bytes, then write them out to disk with slightly different header values.
Scroll would have been extremely fast way to prototype/work with the data, but because I couldn't box the pread data, it exploded my stack; testing with the pwrite version taking references, I was able to write the data, which was one half of the problem.
Hope that's clear?
I think this will generally come up in FFI with large C structs of byte arrays where the api expects a `malloc(sizeof(LargeStruct))` to construct the initial type.

philipc:
I was hoping for an actual real world example. It seems to me that large fixed size arrays are quite uncommon, and even then you can usually use `Vec<u8>` anyway. So the use case is:
- you have a struct containing a large fixed size array
- you need to pass this struct to C with FFI so you can't use `Vec`?

m4b:
It was the other way around, basically:
```c
struct Foo {
  struct Header header;
  uint8_t data [0x8000];
  struct Footer footer; 
}
```
[...]

PR #79, opened 2021-09-14
Pread/Pwrite redefinitions

### LINKED ISSUE #78: Writing into a dynamic container
First of all, I just wanted to say that I really like `scroll`! The context system and type-based parsing and everything works perfectly for my use case.

However, there is one thing that I can't seem to figure out how to do - writing types into a container like a `Vec` where the full size of the type is unknown.

Consider this (admittedly contrived) example:
```rust
pub struct MyType;
impl TryIntoCtx<'_> for MyType {
    type Error = scroll::Error;

    fn try_into_ctx(self, into: &mut [u8], _: ()) -> Result<usize, Self::Error> {
        let offset = &mut 0;

        if /* something that isn't statically determined */
        {
            into.gwrite_with(1u16, scroll::LE)?;
        } else {
            into.gwrite_with(1u32, scroll::LE)?;
        }
        
        Ok(*offset)
    }
}
```

If I pre-allocated a `let mut buf = [0u8; 4]` to cover the u32 case, then I'd be wasting space on the u16 case.

What I would love to do is something like this:
```rust
let mut buf = vec![];
buf.gwrite(MyType, 0)?;
```
where the `Vec<u8>` only allocates as-needed.

Is this something that's possible, potentially with a new container trait?

nickbclifford:
This PR adds a new type, `DynamicBuffer`, that automatically expands when `Pwrite`ing types into it. Details in the doc comments.

It also includes a relaxed definition of the `Pwrite` trait that, while technically an API break, is source-compatible and makes implementing custom containers easier.

Resolves #78.

### DIFF
--- a/src/pwrite.rs
+++ b/src/pwrite.rs
@@ -1,7 +1,6 @@
-use core::ops::{Index, IndexMut, RangeFrom};
 use core::result;
 
-use crate::ctx::{MeasureWith, TryIntoCtx};
+use crate::ctx::TryIntoCtx;
 use crate::error;
 
 /// A very generic, contextual pwrite interface in Rust.
@@ -21,15 +20,12 @@ use crate::error;
 /// should read the documentation of `Pread` first.
 ///
 /// Unless you need to implement your own data store — that is either can't convert to `&[u8]` or
-/// have a data that is not `&[u8]` — you will probably want to implement
+/// have a data that does not expose a `&mut [u8]` — you will probably want to implement
 /// [TryIntoCtx](ctx/trait.TryIntoCtx.html) on your Rust types to be written.
 ///
-pub trait Pwrite<Ctx, E>: Index<usize> + IndexMut<RangeFrom<usize>> + MeasureWith<Ctx>
-where
-    Ctx: Copy,
-    E: From<error::Error>,
-{
-    fn pwrite<N: TryIntoCtx<Ctx, <Self as Index<RangeFrom<usize>>>::Output, Error = E>>(
+pub trait Pwrite<Ctx: Copy, E> {
+    #[inline]
+    fn pwrite<N: TryIntoCtx<Ctx, Self, Error = E>>(
         &mut self,
         n: N,
         offset: usize,
@@ -39,6 +35,7 @@ where
     {
         self.pwrite_with(n, offset, Ctx::default())
     }
+
     /// Write `N` at offset `I` with context `Ctx`
     /// # Example
     /// ```
@@ -46,22 +43,16 @@ where
     /// let mut bytes: [u8; 8] = [0, 0, 0, 0, 0, 0, 0, 0];
     /// bytes.pwrite_with::<u32>(0xbeefbeef, 0, LE).unwrap();
     /// assert_eq!(bytes.pread_with::<u32>(0, LE).unwrap(), 0xbeefbeef);
-    fn pwrite_with<N: TryIntoCtx<Ctx, <Self as Index<RangeFrom<usize>>>::Output, Error = E>>(
+    fn pwrite_with<N: TryIntoCtx<Ctx, Self, Error = E>>(
         &mut self,
         n: N,
         offset: usize,
         ctx: Ctx,
-    ) -> result::Result<usize, E> {
-        let len = self.measure_with(&ctx);
-        if offset >= len {
-            return Err(error::Error::BadOffset(offset).into());
-        }
-        let dst = &mut self[offset..];
-        n.try_into_ctx(dst, ctx)
-    }
+    ) -> result::Result<usize, E>;
+
     /// Write `n` into `self` at `offset`, with a default `Ctx`. Updates the offset.
     #[inline]
-    fn gwrite<N: TryIntoCtx<Ctx, <Self as Index<RangeFrom<usize>>>::Output, Error = E>>(
+    fn gwrite<N: TryIntoCtx<Ctx, Self, Error = E>>(
         &mut self,
         n: N,
         offset: &mut usize,
@@ -72,29 +63,34 @@ where
         let ctx = Ctx::default();
         self.gwrite_with(n, offset, ctx)
     }
+
     /// Write `n` into `self` at `offset`, with the `ctx`. Updates the offset.
     #[inline]
-    fn gwrite_with<N: TryIntoCtx<Ctx, <Self as Index<RangeFrom<usize>>>::Output, Error = E>>(
+    fn gwrite_with<N: TryIntoCtx<Ctx, Self, Error = E>>(
         &mut self,
         n: N,
         offset: &mut usize,
         ctx: Ctx,
     ) -> result::Result<usize, E> {
         let o = *offset;
-        match self.pwrite_with(n, o, ctx) {
-            Ok(size) => {
-                *offset += size;
-                Ok(size)
-            }
-            err => err,
-        }
+        self.pwrite_with(n, o, ctx).map(|size| {
+            *offset += size;
+            size
+        })
     }
 }
 
-impl<
-        Ctx: Copy,
-        E: From<error::Error>,
-        R: ?Sized + Index<usize> + IndexMut<RangeFrom<usize>> + MeasureWith<Ctx>,
-    > Pwrite<Ctx, E> for R
-{
+impl<Ctx: Copy, E: From<error::Error>> Pwrite<Ctx, E> for [u8] {
+    fn pwrite_with<N: TryIntoCtx<Ctx, Self, Error = E>>(
+        &mut self,
+        n: N,
+        offset: usize,
+        ctx: Ctx,
+    ) -> result::Result<usize, E> {
+        if offset >= self.len() {
+            return Err(error::Error::BadOffset(offset).into());
+        }
+        let dst = &mut self[offset..];
+        n.try_into_ctx(dst, ctx)
+    }
 }
--- a/src/pread.rs
+++ b/src/pread.rs
@@ -1,7 +1,6 @@
-use core::ops::{Index, RangeFrom};
 use core::result;
 
-use crate::ctx::{MeasureWith, TryFromCtx};
+use crate::ctx::TryFromCtx;
 use crate::error;
 
 /// A very generic, contextual pread interface in Rust.
@@ -39,14 +38,10 @@ use crate::error;
 /// the resulting type. scroll defaults to `&[u8]` here.
 ///
 /// Unless you need to implement your own data store — that is either can't convert to `&[u8]` or
-/// have a data that is not `&[u8]` — you will probably want to implement
+/// have a data that does not expose a `&[u8]` — you will probably want to implement
 /// [TryFromCtx](ctx/trait.TryFromCtx.html) on your Rust types to be extracted.
 ///
-pub trait Pread<Ctx, E>: Index<usize> + Index<RangeFrom<usize>> + MeasureWith<Ctx>
-where
-    Ctx: Copy,
-    E: From<error::Error>,
-{
+pub trait Pread<Ctx: Copy, E> {
     #[inline]
     /// Reads a value from `self` at `offset` with a default `Ctx`. For the primitive numeric values, this will read at the machine's endianness.
     /// # Example
@@ -54,16 +49,16 @@ where
     /// use scroll::Pread;
     /// let bytes = [0x7fu8; 0x01];
     /// let byte = bytes.pread::<u8>(0).unwrap();
-    fn pread<'a, N: TryFromCtx<'a, Ctx, <Self as Index<RangeFrom<usize>>>::Output, Error = E>>(
+    fn pread<'a, N: TryFromCtx<'a, Ctx, Self, Error = E>>(
         &'a self,
         offset: usize,
     ) -> result::Result<N, E>
     where
-        <Self as Index<RangeFrom<usize>>>::Output: 'a,
         Ctx: Default,
     {
         self.pread_with(offset, Ctx::default())
     }
+
     #[inline]
     /// Reads a value from `self` at `offset` with the given `ctx`
     /// # Example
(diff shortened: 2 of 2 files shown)


nickbclifford:
I was considering adding a `TryIntoCtx<_, DynamicBuffer>` impl to the `Pwrite` derive macro as well, but (no offense intended!) the proc macro doesn't seem to be documented well and I didn't really fully understand the specifics of how it worked, so I ultimately decided not to.

nickbclifford:
Could I get a review @m4b?

src/buffer.rs
```diff
@@ -0,0 +1,294 @@
...
+        self.buffer.index(self.start_offset + index)
+    }
+}
+impl IndexMut<usize> for DynamicBuffer {
+    fn index_mut(&mut self, mut index: usize) -> &mut Self::Output {
+        index += self.start_offset;
+
+        if let Some(diff) = index.checked_sub(self.buffer.len()) {
+            self.buffer
+                .extend(core::iter::repeat(0).take(core::cmp::max(self.alloc_increment, diff + 1)));
```
m4b, line 109: this is a bit hard to read; needs a comment at least

src/buffer.rs
```diff
@@ -0,0 +1,294 @@
...
+        }
+
+        self.write_end = self.write_end.max(index);
```
m4b, line 112: shouldn't this only be updated if the buffer gets resized? if yes, I'd suggest to put it into that block

src/buffer.rs
```diff
@@ -0,0 +1,294 @@
...
+    }
+
+    #[test]
+    fn int_write() {
+        let mut buf = DynamicBuffer::new();
+
+        buf.pwrite_with(0x1234u16, 0, Endian::Little).unwrap();
+        buf.pwrite_with(0x5678i16, 2, Endian::Big).unwrap();
+
+        assert_eq!(buf.get(), [0x34, 0x12, 0x56, 0x78]);
```
m4b, line 209: i haven't thought deeply, but be careful to think about how this test would pass on big endian machine. In other places in scroll there sometimes need to be `cfg(target_endian = "little")` etc.; I think you're ok here but something to think about for a minute, because in my experience easy to make subtle mistake with endianness in tests and it fails years later when someone tries to build it on some big endian machine :)

src/buffer.rs
```diff
@@ -0,0 +1,294 @@
...
+impl<Ctx: Copy, E> Pwrite<Ctx, E> for DynamicBuffer {
+    type Buffer = Self;
+
+    fn pwrite_with<N: TryIntoCtx<Ctx, Self::Buffer, Error = E>>(
+        &mut self,
+        n: N,
+        offset: usize,
+        ctx: Ctx,
+    ) -> Result<usize, E> {
+        self.start_offset += offset;
```
m4b, line 127: hmm, why do you increment start offset by offset then subtract below? might be good to comment this since it isn't immediately clear

src/buffer.rs
```diff
@@ -0,0 +1,294 @@
...
+    /// use scroll::{DynamicBuffer, LE, Pwrite};
+    /// let mut buf = DynamicBuffer::new();
+    ///
+    /// buf.pwrite_with(2u32, 0, LE).unwrap();
+    /// assert_eq!(buf.get(), [2, 0, 0, 0]);
+    /// ```
+    pub fn get(&self) -> &[u8] {
+        &self.buffer[..=self.write_end]
+    }
+}
```
m4b, line 94: I'm on fence about adding:
```rust
pub fn into_inner(self) -> Vec<u8> {
  self.buffer
}
```

on one hand it's nice and typical api of this sort; on other hand it forever forces the implementation to either use a Vec<u8> internally or create one on the fly if the user calls this method. It's probably fine?

src/buffer.rs
```diff
@@ -0,0 +1,294 @@
...
+/// we cannot create a custom slice type that performs dynamic expansions under the hood.
+///
+/// As a result, we need to expose a separate writing type that can track offsets and calculate when to allocate.
+///
+/// However, if your `TryIntoCtx` impls don't use any special slice APIs and just use `Pwrite` and/or
+/// basic indexing, it's extremely easy to migrate! Just add the `DynamicBuffer` generic type.
+pub struct DynamicBuffer {
+    buffer: Vec<u8>,
+    alloc_increment: usize,
+    start_offset: usize,
```
m4b, line 63: reading through this it isn't exactly clear why start_offset is for, and why the extra bookkeeping associated with it; i'm sure it is required, but might be nice to have some comments explaining it's use, etc.

src/pwrite.rs
```diff
@@ -24,12 +24,11 @@ use crate::error;
 /// have a data that is not `&[u8]` — you will probably want to implement
 /// [TryIntoCtx](ctx/trait.TryIntoCtx.html) on your Rust types to be written.
 ///
-pub trait Pwrite<Ctx, E>: Index<usize> + IndexMut<RangeFrom<usize>> + MeasureWith<Ctx>
-where
-    Ctx: Copy,
-    E: From<error::Error>,
-{
-    fn pwrite<N: TryIntoCtx<Ctx, <Self as Index<RangeFrom<usize>>>::Output, Error = E>>(
+pub trait Pwrite<Ctx: Copy, E> {
```
m4b, line 27: Kind of talking to myself: So re my worries about dispatch ambiguity, i think it's not possilble; since implementors (e.g., blanket impl + DynamicBuffer) implement Pwrite, etc., and so disaptch is always `thing_that_implements_pwrite.pwrite` so there won't be any dispatch ambiguity, so yea that's a worry I don't need to have.

src/pwrite.rs
```diff
@@ -24,12 +24,11 @@ use crate::error;
...
+    /// The type used by implementors of `TryIntoCtx` to write into your data store.
+    type Buffer: ?Sized;
```
m4b, line 29: so this part is peculiar to me; i have to think about it more; specifically, I'm wondering why this associated type is required; in e.g., your impl Buffer is Self, and pwrite requires basically `TryIntoCtx<Ctx, Self::Buffer, Error>`, which means `TryIntoCtx<Ctx, Self, Error>` so why wouldn't the associated type be dropped and we just pass Self there always?

I'm sure there's a good reason, but it would be nice if possible not to introduce the associated type, though if it is requried I think it's fine. I don't know of anyone putting explicit `Pwrite` bounds on their generic types (it shoudl only require `TryIntoCtx` bounds in order for `Pwrite` to work.

Quite an interesting solution to the problem you've done!

src/pwrite.rs
```diff
@@ -39,29 +38,24 @@ where
...
         ctx: Ctx,
-    ) -> result::Result<usize, E> {
-        let len = self.measure_with(&ctx);
-        if offset >= len {
-            return Err(error::Error::BadOffset(offset).into());
-        }
-        let dst = &mut self[offset..];
-        n.try_into_ctx(dst, ctx)
-    }
+    ) -> result::Result<usize, E>;
```
m4b, line 54: so I don't like that this default impl is removed, but then I forgot there is a blanket impl of Pwrite below anyway, which is empty, because of this default impl; and this just gets moved there. The logic is a bit more separated, but it makes custom container types more flexible, which is quite cool.

src/pwrite.rs
```diff
@@ -97,4 +92,19 @@ impl<
         R: ?Sized + Index<usize> + IndexMut<RangeFrom<usize>> + MeasureWith<Ctx>,
```
m4b, line 92: one thing that's puzzling me is how this blanket impl doesn't overlap the implementation of Pwrite for DynamicBuffer? (i.e., violates single implementation rule)

src/lib.rs
```diff
@@ -229,6 +231,8 @@ mod lesser;
 mod pread;
 mod pwrite;
 
+#[cfg(feature = "std")]
+pub use crate::buffer::*;
```
m4b, line 235: what's funny in scroll prehistory is it had a module named `buffer` which did something similar (but it was just wrapping `&[u8]` basically, and wasn't dynamic), before I realized I could do everything generically with existing types.

src/lib.rs
```diff
@@ -219,6 +219,8 @@ pub use scroll_derive::{IOread, IOwrite, Pread, Pwrite, SizeWith};
 #[cfg(feature = "std")]
 extern crate core;
 
+#[cfg(feature = "std")]
+mod buffer;
```
m4b, line 223: we may want to add a feature flag for dynamic buffers, not sure.

src/buffer.rs
```diff
@@ -0,0 +1,294 @@
+use crate::ctx::TryIntoCtx;
```
m4b, line 1: I think this entire module could actually be a third party crate; it is a self-defined type which implements a foreign trait, so ti won't violate orphan rules. (assuming we merge your Pwrite/Pread changes).

m4b (commented):
This is pretty awesome! I'm quite surprised that tests pass, without any changes, honestly, although it is technically a breaking change (which is fine).

So, some thoughts:
1. I think the associated type Buffer is unavoidable? (which is fine if so, this is the technically breaking change)
2. Should Pread be updated (I don't think your DynamicBuffer works with pread? It might be nice from a symmetric viewpoint to get that working, but I don't think its' necessary)
3. There's also the argument to be made that this type could be in a third party crate (assuming the Pwrite changes go in)

The one thing I don't like is the asymmetry now between Pwrite/Pread; one of them has an associated type and the other doesn't. Is there a way to rectify this, you think?

Also what are you thoughts on DynamicBuffer being in a third party crate? (I think it's possible, I don't see an orphan rule violation anywhere). This would also punt on deciding on whether it should be under a feature flag (I generally don't like feature flags if can be avoided).

Lastly, w.r.t. the scroll_derive issues, I have to think about it, but I don't think we should generate extra implementations for DynamicBuffer in the default derive, since for my guess is majority of usecases won't require a DynamicBuffer, and it's just extra code bloat having them compile it.

Anyway, I don't see any _immediate_ issues preventing this from being merged; it's quite cool, and thanks for the PR; as I said, will need to be a little patient getting this in, don't have the most time right now, but I don't see any major blocking issues, however I would like to play with it a bit see if something comes up.

Anyway, feel free to add your thoughts here, etc. :) really great stuff :)

m4b:
also it doesn't look like there's any CI anymore on scroll which is terrifying? (it used to run on jenkins, but it is probably dead?)

src/pwrite.rs
nickbclifford: That's a good point, and I think you might be right. I added an associated type to make sure that the blanket implementations that already existed wouldn't change behavior, but if we change that to just a `impl [u8]` or something, then that shouldn't be a problem and we can get rid of it.

src/pwrite.rs
nickbclifford: `DynamicBuffer` doesn't have a range index or implement `MeasureWith`.

nickbclifford:
I'd be okay with moving `DynamicBuffer` into a separate crate, I believe all that would be necessary are the changes for Pwrite.

However, I do agree that the asymmetry with Pread is annoying, so I'll change that as well.

nickbclifford:
Let me know what you think about the trait changes, and I'd be happy to move the buffer stuff out of this PR and change scope.

m4b:
I have not forgotten about this! I will try to review tomorrow, thanks for your patience; always feel free to ping me if you aren't hearing from me, also :)

nickbclifford:
@m4b Any thoughts about merging soon?

m4b:
@nickbclifford ok, so here's what we should do:

1. if you can squash all your commits to single, and force push, that will make for a nice final review
2. I am inclined to merge this breaking change, without the `type Buffer` associated type; although this prevents some uses of const buffers, this may not ever matter, it would be an interesting case study to examine the assembly output to see if rustc is able to do anything with array vs slice

Unrelated to your patch, but you probably want a 0.11 crates release at some point:
1. i will likely remove the `usize` impl for `pread` and `pwrite` #56 
2. i will possibly look at making `pwrite` take `&` references, which is something i wanted for some time, though it may be a huge change, which is related to #51 and was the PR that never got merged here #47 

I would really want to revisit `pwrite` taking `&` before releasing 0.11, which at this stage would be effectively a beta for scroll 1.0. Before 1.0 i think pwrite taking references is good future proof design.

So in summary, if you can squash your changes, I think i'm ok to merge, and then i'd like to see at least 1. (but probably not 2., i'm not sure who would do it and i don't have time right now) before releasing 0.11

Does that sound reasonable? And once again, thanks for your patience here :)

nickbclifford:
Yep, sounds all good to me!

m4b (approved):
LGTM! So thinking more about doing:

```rust
impl<Ctx: Copy, E: From<error::Error>, const LEN: usize> Pwrite<Ctx, E> for [u8; LEN] {
```
the problem here is that we want the buffer we write _into_ to be at most the size of the type; but we don't know the type until `pwrite` is called, e.g.:

```rust
fn pwrite_with<N: TryIntoCtx<Ctx, Self, Error = E>>(
```
which means we'd want something like:

```rust
fn pwrite_with<N: TryIntoCtx<Ctx, [u8; size_of::<N>()], Error = E>>(
```

which means the pwrite dispatches _from_ a buffer of some size N, _into_ a buffer that can contain the type.

If we just dispatch to the a slice, the value of const generics is effectively nil, since it is just a check on some constant `N` instead of `self.len()`.

What is interesting here is I don't think this pattern is expressible in scroll as defined, nor will an associated `type Buffer` help, since it's size is given in the impl block, but the size is effectively only known at `pwrite` definition time.

I think the only solution here would be to add another generic type to `Pwrite` which is a marker for the written to buffer? I'm not sure, have to think about it, but I've convinced myself that adding `type Buffer` adds no real benefit, other than allowing to dispatch _from_ const arrays to slices, which has minimal benefit.

Anyway, great PR, great discussion, and thanks for your patience!

PR #84, opened 2023-08-11
Minor linting, warning fixes

nyurik:
* Optimize/cleanup use statements
* Do not expose `scroll_derive` as a feature in Cargo.toml
* Minor clippy lints

- [x] There is a FIXME in the code that needs to be addressed before merging

The format args inlining was moved to #91

### DIFF
--- a/src/ctx.rs
+++ b/src/ctx.rs
@@ -181,11 +181,8 @@
 //! ```
 
 use core::mem::size_of;
-use core::mem::transmute;
 use core::ptr::copy_nonoverlapping;
-use core::result;
-use core::str;
-
+use core::{result, str};
 #[cfg(feature = "std")]
 use std::ffi::{CStr, CString};
 
@@ -240,18 +237,14 @@ impl Default for StrCtx {
 
 impl StrCtx {
     pub fn len(&self) -> usize {
-        match *self {
+        match self {
             StrCtx::Delimiter(_) | StrCtx::DelimiterUntil(_, _) => 1,
             StrCtx::Length(_) => 0,
         }
     }
 
     pub fn is_empty(&self) -> bool {
-        if let StrCtx::Length(_) = *self {
-            true
-        } else {
-            false
-        }
+        matches!(self, StrCtx::Length(_))
     }
 }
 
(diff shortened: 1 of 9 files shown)


m4b:
@nyurik i'm very sorry I missed/dropped this PR, I don't know if you're still interested, but getting nostd tests passing is definitely something worthwhile.

If you are still interested, I think splitting out just getting nostd tests passing will be useful (and easier to review), and then if you like, fixing various lints and other things as a separate PR.

nyurik:
@m4b thx for review, i submitted #89 and rebased this PR on top of it

nyurik:
@m4b hi, is there anything else that should be fixed in this PR? Thx!

src/ctx.rs
```diff
@@ -625,13 +619,7 @@ macro_rules! into_ctx_float_impl {
...
             fn into_ctx(self, dst: &mut [u8], le: Endian) {
                 assert!(dst.len() >= $size);
-                write_into!(
-                    signed_to_unsigned!($typ),
-                    $size,
-                    transmute::<$typ, signed_to_unsigned!($typ)>(self),
-                    dst,
-                    le
-                );
+                write_into!(signed_to_unsigned!($typ), $size, self.to_bits(), dst, le);
```
m4b, line 622: some of this is rolling up fixes from the other PR that fixes unsafe stuff, i'm not sure it belongs in linting issue (unless clippy auto fixes this?)

tests/api.rs
```diff
@@ -1,14 +1,7 @@
-// this exists primarily to test various API usages of scroll; e.g., must compile
```
m4b, line 1: i don't think this comment needed to be deleted

src/ctx.rs
```diff
@@ -407,13 +400,14 @@ macro_rules! signed_to_unsigned {
 
 macro_rules! write_into {
     ($typ:ty, $size:expr, $n:expr, $dst:expr, $endian:expr) => {{
+        assert!($dst.len() >= $size);
```
m4b, line 403: e.g., here too, this moving out of the unsafe block probably conflicts with other pr, but it's fine i think

m4b (commented):
some minor nits

src/ctx.rs
nyurik: I moved it out of the unsafe block on purpose - to my knowledge, the unsafe should be as small as possible - limited to just the unsafe aspect of the code. I know it conflicts with the other PR, but I think we can easily fix those.  If ok, I would prefer to keep it, but up to you?

src/ctx.rs
nyurik: yes, this is what clippy fixes now

tests/api.rs
nyurik: thx, reverted

nyurik:
@m4b thx, replied.  Could you take a look at https://github.com/m4b/scroll/pull/84/files#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759R273-R274 -- the test in src/lib.rs -- is `#![allow(overflowing_literals)]` still needed?

src/ctx.rs
m4b: sounds good

m4b:
You could try removing the allow and see what happens? I can't recall what it's there, it was probably added years ago for some reason lost to time :)

nyurik:
@m4b good point, removed, seems fine :)

m4b (approved):
thank you for making these changes, it's a bit thankless work so i'm thanking you :)

nyurik:
its not thankless if its gets merged :)  Thx for working on this!

PR #88, opened 2023-11-05
Fix clippy warnings.

Frostie314159:
This also removes multiple uses of transmute, which have safe equivalents. All tests pass and clippy is happy.

### DIFF
--- a/src/ctx.rs
+++ b/src/ctx.rs
@@ -191,7 +191,6 @@ use std::ffi::{CStr, CString};
 use crate::endian::Endian;
 use crate::error;
 use crate::Pread;
-use crate::Pwrite;
 
 /// A trait for measuring how large something is; for a byte sequence, it will be its length.
 pub trait MeasureWith<Ctx> {
@@ -847,20 +846,6 @@ impl<'a, const N: usize> TryFromCtx<'a> for [u8; N] {
     fn try_from_ctx(from: &'a [u8], _ctx: ()) -> Result<(Self, usize), Self::Error> {
         // Unwrap is OK, since pread_with already asserts the length.
         Ok((from.pread_with::<&'a [u8]>(0, N)?.try_into().unwrap(), N))
-    }
-}
-impl<const N: usize> TryIntoCtx for [u8; N] {
-    type Error = error::Error;
-    #[inline]
-    fn try_into_ctx(self, from: &mut [u8], _ctx: ()) -> Result<usize, Self::Error> {
-        from.pwrite(self.as_slice(), 0)
-    }
-}
-impl<'a, const N: usize> TryIntoCtx for &'a [u8; N] {
-    type Error = error::Error;
-    #[inline]
-    fn try_into_ctx(self, from: &mut [u8], ctx: ()) -> Result<usize, Self::Error> {
-        (*self).try_into_ctx(from, ctx)
     }
 }
 
--- a/src/ctx.rs
+++ b/src/ctx.rs
@@ -718,7 +718,7 @@ impl<'a> TryIntoCtx for &'a [u8] {
         let src_len = self.len() as isize;
         let dst_len = dst.len() as isize;
         // if src_len < 0 || dst_len < 0 || offset < 0 {
-        //     return Err(error::Error::BadOffset(format!("requested operation has negative casts: src len: {} dst len: {} offset: {}", src_len, dst_len, offset)).into())
+        //     return Err(error::Error::BadOffset(format!("requested operation has negative casts: src len: {src_len} dst len: {dst_len} offset: {offset}")).into())
         // }
         if src_len > dst_len {
             Err(error::Error::TooBig {
--- a/src/ctx.rs
+++ b/src/ctx.rs
@@ -853,6 +853,7 @@ impl TryIntoCtx for CString {
 }
 impl<'a, const N: usize> TryFromCtx<'a> for [u8; N] {
     type Error = error::Error;
+    #[inline]
     fn try_from_ctx(from: &'a [u8], _ctx: ()) -> Result<(Self, usize), Self::Error> {
         // Unwrap is OK, since pread_with already asserts the length.
         Ok((from.pread_with::<&'a [u8]>(0, N)?.try_into().unwrap(), N))
@@ -860,12 +861,14 @@ impl<'a, const N: usize> TryFromCtx<'a> for [u8; N] {
 }
 impl<const N: usize> TryIntoCtx for [u8; N] {
     type Error = error::Error;
+    #[inline]
     fn try_into_ctx(self, from: &mut [u8], _ctx: ()) -> Result<usize, Self::Error> {
         from.pwrite(self.as_slice(), 0)
     }
 }
 impl<'a, const N: usize> TryIntoCtx for &'a [u8; N] {
     type Error = error::Error;
+    #[inline]
     fn try_into_ctx(self, from: &mut [u8], ctx: ()) -> Result<usize, Self::Error> {
         (*self).try_into_ctx(from, ctx)
     }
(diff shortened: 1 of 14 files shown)


src/ctx.rs
```diff
@@ -621,13 +619,7 @@ macro_rules! into_ctx_float_impl {
             #[inline]
             fn into_ctx(self, dst: &mut [u8], le: Endian) {
                 assert!(dst.len() >= $size);
-                write_into!(
-                    signed_to_unsigned!($typ),
-                    $size,
-                    transmute::<$typ, signed_to_unsigned!($typ)>(self),
```
m4b, line 627: just want to be sure this doesn't affect semantics here, the `signed_to_unsigned!($typ)` in the transmute param is ok to replace with just to_bits(), yes?  (I haven't reviewed this code in literally years and years, so my memory for why this stuff is even here is foggy)

m4b (approved):
this looks like a great initial cleanup, thank you so much! only question was on the one transmute, but it looks fine to me. otherwise i think we have to merge your other PR and fix ci in it before we merge this one?

src/ctx.rs
Frostie314159: The [docs](https://doc.rust-lang.org/std/primitive.f32.html#method.to_bits) for this method show, that the unsigned version is always returned.

src/ctx.rs
```diff
@@ -849,6 +841,28 @@ impl TryIntoCtx for CString {
         self.as_c_str().try_into_ctx(dst, ())
     }
 }
+impl<'a, const N: usize> TryFromCtx<'a> for [u8; N] {
```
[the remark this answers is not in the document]
Frostie314159: I'm gonna revert those commits, since they weren't intended to be here in the first place.

nyurik:
I fixed the CI - MSRV in #90

nyurik:
I fixed the CI - MSRV in #90

src/ctx.rs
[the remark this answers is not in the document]
Frostie314159: Fixed in latest commit.

m4b:
so if you're having trouble fixing this, you can do:
```
git checkout master
git pull
git checkout clippy-fix # assumes this is your branch name
git rebase master
# you will likely have to manually fix the conflict markers (choose the 0x4a case)
git add src/lib.rs
git rebase --continue
git push --force # force push the rebased change to your branch 
```

nyurik:
@m4b and @Frostie314159 I created a PR for this PR - fixing merge conflicts. It should be merged in before merging this PR - https://github.com/Frostie314159/scroll/pull/1

m4b:
@nyurik it still says there is a merge conflict?

m4b:
Also, i'm not sure what's going on, but i'm having trouble reviewing the "final changes" of this PR; when i choose to review all commits, only 3 files are changed, and some changes that were supposedly removed (the try_from_ctx impl for `[u8; N]`) is still present, and all the transmute and safety fixes are missing. :confused: 
review_screenshot

I'd expect reviewing all changes to show me what i'm going to merge, maybe i haven't done this in a while and i my memory doesn't serve, but i don't remember github presenting stacks of commits in this manner? (i remember being able to view what he entire set of changes introduced, which is what i want to see, and which is what would be merged into master)

nyurik:
@m4b those fixes were also done and merged in by my earlier PR.  This PR only has the stuff that my (also clippy-inspired PR) did not have. The diff tool (add `/compare` to your repo) will help us explore this:

Main branch has two commits: f5beda1c9 and 30f49c7bb. The second commit is mostly irrelevant (format inlining), so comparing with it is pointless because it will just increase the noise. The first commit on the other hand is the one that contains a lot of the changes that were also done by this commit.

You can see it by comparing it with 62c5fcf90325f: 
 https://github.com/m4b/scroll/compare/f5beda1c9..62c5fcf -- this is the commit right before my PR was merged into the @Frostie314159 fork.

Now the very next commit - caf60bb: 
* this is what changed between @Frostie314159 commits: https://github.com/m4b/scroll/compare/62c5fcf..caf60bb
* this is the new diff with the primary branch https://github.com/m4b/scroll/compare/f5beda1c9..caf60bb

Next is the merge commit --  eac19ef -- merge commit is a noop for us, so same results as before
* this is what changed between @Frostie314159 commits: https://github.com/m4b/scroll/compare/caf60bb..eac19ef
* this is the new diff with the primary branch https://github.com/m4b/scroll/compare/f5beda1c9..eac19ef

So in short - all transmute fixes are already in the main branch.

nyurik:
P.S. the conflict is tiny - I fixed it in https://github.com/Frostie314159/scroll/pull/2 -- this is again just a simple PR that was created with `git merge upstream/master` and fixing conflicts.

-- this PR modifies two lines: [this](https://github.com/m4b/scroll/pull/88/files#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759R388) and [this](https://github.com/m4b/scroll/pull/88/files#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759R397) - both in `src/lib.rs` - and my last PR modifies `println!("1 {:?}", &hello_world);` which is right above those lines. So git requires manual check when two adjacent lines are modified by two separate commits.

Frostie314159:
Thank you so much!

src/lib.rs
```diff
@@ -486,9 +483,9 @@ mod tests {
         let bytes_from: [u8; 8] = [1, 2, 3, 4, 5, 6, 7, 8];
         let bytes_to = &mut bytes_to[..];
         let bytes_from = &bytes_from[..];
-        for i in 0..bytes_from.len() {
-            bytes_to[i] = bytes_from.pread(i).unwrap();
-        }
+        bytes_to.iter_mut().enumerate().for_each(|(i, item)| {
+            *item = bytes_from.pread(i).unwrap();
+        });
```
nyurik, line 488: i think this is a bit hard to read - iterating over the "target" as mutable.  I would suggest a simpler variant (i didn't test it, might not even compile) -- iterate over source values, pushing them into the output array.


```suggestion
        for (idx, value) in bytes_from.iter().enumerate() {
            bytes_to[i] = value.unwrap();
        });
```

src/lib.rs
```diff
@@ -580,10 +577,10 @@ mod tests {
         let bytes_from: [u8; 8] = [1, 2, 3, 4, 5, 6, 7, 8];
         let bytes_to = &mut bytes_to[..];
         let bytes_from = &bytes_from[..];
-        let mut offset = &mut 0;
-        for i in 0..bytes_from.len() {
-            bytes_to[i] = bytes_from.gread(&mut offset).unwrap();
-        }
+        let offset = &mut 0;
+        bytes_to.iter_mut().for_each(|item| {
```
nyurik, line 581: same as above

tests/api.rs
```diff
@@ -136,6 +137,11 @@ impl<'a> Segments<'a> {
         Ok(sections)
     }
 }
+impl<'a> Default for Segments<'a> {
```
nyurik, line 140: Why is this needed?  You can simply add `#[derive(Default)]` on the `Segments` struct, and make `new()` call `Self::default()`.  Also, I am not too certain if both `default()` and `new()` are needed. I think cleaner interface would be to have just one. This way if later you need `new(some_init_value)`, it can be easily added.

src/lib.rs
nyurik: although this might need to be different - not certain about this one

Frostie314159: This was the clippy suggestion.

tests/api.rs
Frostie314159: Good idea, I can get to this tomorrow.

m4b:
i believe all these fixes have been merged, if not we should do this in another PR since this has conflicts, etc.

PR #89, opened 2023-11-05
Add nostd tests

nyurik:
This uses MSRV fix #90

* Fix `cargo test --no-default-features` to pass by making std-only tests/docs conditional (added to CI)
* Added `Error::Custom(&'static str)` for no-std mode

There are a few `FIXME` in the code where things seemed suspicious

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -56,3 +56,4 @@ jobs:
         cargo fmt -- --check
     - run: cargo build
     - run: cargo test
+    - run: cargo test --no-default-features
--- a/src/error.rs
+++ b/src/error.rs
@@ -23,6 +23,9 @@ pub enum Error {
     #[cfg(feature = "std")]
     /// A custom Scroll error for reporting messages to clients
     Custom(String),
+    #[cfg(not(feature = "std"))]
+    /// A custom static Scroll error for reporting messages to clients
+    Custom(&'static str),
     #[cfg(feature = "std")]
     /// Returned when IO based errors are encountered
     IO(io::Error),
@@ -72,6 +75,10 @@ impl Display for Error {
             #[cfg(feature = "std")]
             Error::Custom(ref msg) => {
                 write!(fmt, "{}", msg)
+            }
+            #[cfg(not(feature = "std"))]
+            Error::Custom(msg) => {
+                write!(fmt, "{msg}")
             }
             #[cfg(feature = "std")]
             Error::IO(ref err) => {
--- a/src/error.rs
+++ b/src/error.rs
@@ -20,14 +20,13 @@ pub enum Error {
         size: usize,
         msg: &'static str,
     },
-    #[cfg(feature = "std")]
     /// A custom Scroll error for reporting messages to clients
-    Custom(String),
-    #[cfg(not(feature = "std"))]
-    /// A custom static Scroll error for reporting messages to clients
-    Custom(&'static str),
     #[cfg(feature = "std")]
+    Custom(String),
+    /// A custom static Scroll error for reporting messages to clients
+    CustomStatic(&'static str),
     /// Returned when IO based errors are encountered
+    #[cfg(feature = "std")]
     IO(io::Error),
 }
 
@@ -39,6 +38,7 @@ impl error::Error for Error {
             Error::BadOffset(_) => "BadOffset",
             Error::BadInput { .. } => "BadInput",
             Error::Custom(_) => "Custom",
+            Error::CustomStatic(_) => "Custom", // Both Custom and CustomStatic are custom errors
             Error::IO(_) => "IO",
         }
     }
@@ -48,6 +48,7 @@ impl error::Error for Error {
             Error::BadOffset(_) => None,
             Error::BadInput { .. } => None,
             Error::Custom(_) => None,
+            Error::CustomStatic(_) => None,
             Error::IO(ref io) => io.source(),
         }
     }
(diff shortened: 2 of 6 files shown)


src/error.rs
```diff
@@ -23,6 +23,9 @@ pub enum Error {
     #[cfg(feature = "std")]
     /// A custom Scroll error for reporting messages to clients
     Custom(String),
+    #[cfg(not(feature = "std"))]
+    /// A custom static Scroll error for reporting messages to clients
+    Custom(&'static str),
```
m4b, line 28: I'd like to understand why this change occurred; ideally fixing the tests to run on no std shouldn't require changing the error type here.

If we are required to fix this, I would prefer to name this variant Static and have it always present, if possible; ideally the std flag is additive and doesn't change the meaning of the enum variant when active/not active. Alternatively we could explore Cow and alloc feature on core, but probably a static variant is best

Cargo.toml
```diff
@@ -10,6 +10,7 @@ license = "MIT"
 documentation = "https://docs.rs/scroll"
 description = "A suite of powerful, extensible, generic, endian-aware Read/Write traits for byte buffers"
 include = ["src/**/*", "Cargo.toml", "LICENSE", "README.md"]
+rust-version = "1.63"
```
m4b, line 13: remove

.github/workflows/main.yml
```diff
@@ -13,9 +13,9 @@ jobs:
       matrix:
         build: [MSRV, stable, nightly, macos, win32, win64]
         include:
-          - build: MSRV # Minimum supported Rust version
+          - build: MSRV # Minimum supported Rust version, ensure it matches the rust-version in Cargo.toml
             os: ubuntu-latest
-            rust: 1.62.1
+            rust: 1.63
```
m4b, line 18: dup

src/lib.rs
```diff
@@ -612,12 +623,18 @@ mod tests {
         let res = b.gread_with::<&str>(offset, StrCtx::Length(3));
         assert!(res.is_err());
         *offset = 0;
-        let astring: [u8; 3] = [0x45, 042, 0x44];
+        // FIXME: 0x2A used to be 042 -- a weird way to declare a number, probably a bug
```
m4b, line 626: this was definitely a typo/bug :) should have been 0x42

m4b (changes requested):

src/error.rs
nyurik: @m4b neither the `Custom` nor `Static` errors are required - they are both used for user errors from what i can tell, but in the nostd env, it would not be possible to have a custom error for users - so adding a static one is the simplest way to solve it.  I think `CustomStatic` is a clearer error type - what do you think?

.github/workflows/main.yml
nyurik: this is part of the #90 on which this PR was based.

src/lib.rs
nyurik: thx, fixed

Cargo.toml
nyurik: this is part of the #90 on which this PR was based.

src/error.rs
m4b: that seems ok, yea; honestly i would prefer Cow (for this exact) usecase but it only exists in alloc, so we'd need an alloc flag, and then in any baremetal system you'd still have the same problem, so CustomStatic seems fine I guess

nyurik: thx, so any changes needed for this one still, or is it good to merge?

src/error.rs
```diff
@@ -20,11 +20,13 @@ pub enum Error {
         size: usize,
         msg: &'static str,
     },
-    #[cfg(feature = "std")]
     /// A custom Scroll error for reporting messages to clients
-    Custom(String),
     #[cfg(feature = "std")]
+    Custom(String),
+    /// A custom static Scroll error for reporting messages to clients
+    CustomStatic(&'static str),
```
m4b, line 27: ok sorry to go back and forth, but can you remove these changes (adding CustomStatic); i thought it would be required to get the tests to pass, but a change like this (which is a breaking change) would require clients to update on the additional matches. Because this is just for fixing tests I don't think it should be included in this PR.

Additionally, it's unclear to me why we need this, since technically in a no-std environment the user of scroll could return BadInput which also has a static str (they could set size to 0 if they wanted).

Once the static str stuff is walked out we can merge this immediately, sorry for the busy work and appreciate you getting these changes in, but adding a new variant should be discussed separately I think (and isn't required for this, so let's remove it)

m4b (changes requested):
just remove CustomStatic for now, we can consider it in a follow up/additional patch, thanks, otherwise this is good to go

src/error.rs
nyurik: ok, np, removed

src/error.rs
```diff
@@ -20,11 +20,11 @@ pub enum Error {
         size: usize,
         msg: &'static str,
     },
-    #[cfg(feature = "std")]
     /// A custom Scroll error for reporting messages to clients
-    Custom(String),
     #[cfg(feature = "std")]
+    Custom(String),
     /// Returned when IO based errors are encountered
+    #[cfg(feature = "std")]
```
m4b, line 27: i just realized this got added; this is technically a breaking change, but nostd could never have worked with this, iiuc?

m4b (approved):

src/error.rs
nyurik: correct, `io` does not exist without `std`, so technically it is not a breaking change :)

PR #90, opened 2023-11-05
Fix CI errors due to incorrect MSRV

nyurik:
* Set minimum rust version to 1.63 due to rayon-core MSRV requirements, and add a matching `rust-version` in the Cargo.toml


m4b (approved):
thank you!

Cargo.toml
```diff
@@ -10,6 +10,7 @@ license = "MIT"
 documentation = "https://docs.rs/scroll"
 description = "A suite of powerful, extensible, generic, endian-aware Read/Write traits for byte buffers"
 include = ["src/**/*", "Cargo.toml", "LICENSE", "README.md"]
+rust-version = "1.63"
```
m4b, line 13: argh actually i didn't want to bump the msrv here because we don't have an actual requirement in scroll itself (only the dev deps)

PR #91, opened 2023-11-06
Inline format args (better perf)

nyurik:
This is a subset of the original #84 

Make code a bit easier to read, and also, per rust issue 112156, there is a 6% perf hit when using refs with format, so this PR also fixes that.

See https://github.com/rust-lang/rust/issues/112156

### DIFF
--- a/examples/data_ctx.rs
+++ b/examples/data_ctx.rs
@@ -20,5 +20,5 @@ fn main() {
     let data = bytes.pread_with::<Data>(0, BE).unwrap();
     assert_eq!(data.id, 0x01020304);
     assert_eq!(data.name.to_string(), "UserName".to_string());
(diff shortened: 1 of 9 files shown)


nyurik:
@m4b can this be merged too? It is another fairly simple cleanup

m4b:
i don't think it matters for tests (or does it) but which rust version did fmt strings get the `{foo}` syntax? we are set to 1.63 at the moment

nyurik:
@m4b format inline was in 1.58 - https://releases.rs/docs/1.58.0/

m4b (approved):
great thanks for looking it up, i does matter actually since error.rs (which would be compiled by downstream crates) now uses this feature

PR #92, opened 2023-11-18
pwrite: allow writes to empty slice

baloo:
`slice[len..]` is a valid subslice, this is going to generate an empty subslice, but this won't be considered out of range.

https://play.rust-lang.org/?version=stable&mode=release&edition=2021&gist=31f38d269157a3f6c6ac7019d912456f

cc @RaitoBezarius


baloo:
Duplicate of #81

PR #93, opened 2023-11-20
Provide Try{Into|From}Ctx impls for fixed byte array

nyurik:
This is a squashed #86.  I do not think this should be merged as is because it has a few reverts and a few other unrelated things.

### DIFF
--- a/tests/api.rs
+++ b/tests/api.rs
@@ -1,3 +1,4 @@
+#![allow(clippy::disallowed_names)]
 // this exists primarily to test various API usages of scroll; e.g., must compile
 
 use std::ops::{Deref, DerefMut};
@@ -134,6 +135,11 @@ impl<'a> Segments<'a> {
             sections.push(segment.sections()?);
         }
         Ok(sections)
+    }
+}
+impl<'a> Default for Segments<'a> {
+    fn default() -> Self {
+        Self::new()
     }
 }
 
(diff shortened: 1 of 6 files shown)


nyurik:
I just pushed a new change to this PR - which tries to fix a few merge conflicts.

@m4b - I do not think the macro rules should be changed. @Frostie314159 - did you intend to change those?

tests/api.rs
```diff
@@ -1,3 +1,4 @@
+#![allow(clippy::disallowed_names)]
```
nyurik, line 1: i'm not sure this is needed.

m4b:
I'm sorry I'm getting very confused with what's going on with all these PRs; I think at this stage we have to have the simple PR with a single commit that just adds the impls for fixed byte arrays; I don't want to review more diffs with unrelated changes, clippy changes, etc.

nyurik:
@m4b agree - I also proposed this in #86. I created this PR only to help @Frostie314159 to see what his PR actually changes compared with the main branch.  There is no need to merge it.

m4b:
i see, thank you @nyurik , i'm going to go to sleep now, it's late for me, thanks for both of your hard work and PRs, much appreciated :)

Frostie314159:
Thanks for the help!

m4b:
this is done in another pr

Issue #53, opened 2018-05-02
Generated Pread code is large/slow


jrmuizel:
The following rust code:
```rust
#[derive(Pread, Pwrite, IOread, IOwrite)]
struct Data {
    id: u32,
    x: u32,
    y: u32,
    z1: u32,
}

#[inline(never)]
fn make(input: &[u8]) -> Result<Data, scroll::Error> {
    let data = input.pread_with::<Data>(0, LE);
    data
}

fn main (){
    let bytes = Vec::new();
    make(&bytes).unwrap();
}
```

compiles to:

```assembly
pread_example::make:
 push    rbp
 mov     rbp, rsp
 mov     r8d, 1
 test    rdx, rdx
 je      LBB19_4
 cmp     rdx, 4
 jae     LBB19_5
 mov     esi, 4
 xor     eax, eax
LBB19_3:
 xor     ecx, ecx
 jmp     LBB19_9
LBB19_4:
 xor     esi, esi
 mov     eax, 1
 xor     ecx, ecx
 jmp     LBB19_8
LBB19_5:
 cmp     rdx, 4
 jne     LBB19_11
 xor     ecx, ecx
 mov     esi, 4
LBB19_7:
 mov     eax, 1
LBB19_8:
LBB19_9:
LBB19_10:
 or      rax, rcx
 mov     dword, ptr, [rdi], r8d
 mov     dword, ptr, [rdi, +, 4], r9d
 mov     qword, ptr, [rdi, +, 8], rax
 mov     qword, ptr, [rdi, +, 16], rsi
 mov     qword, ptr, [rdi, +, 24], rdx
 pop     rbp
 ret
LBB19_11:
 lea     r9, [rdx, -, 4]
 cmp     r9, 4
 jae     LBB19_13
LBB19_12:
 mov     esi, 4
 xor     eax, eax
 xor     ecx, ecx
 mov     rdx, r9
 jmp     LBB19_9
LBB19_13:
 cmp     rdx, 9
 jb      LBB19_18
 lea     r9, [rdx, -, 8]
 cmp     r9, 4
 jb      LBB19_12
 cmp     rdx, 13
 jb      LBB19_19
 add     rdx, -12
 xor     eax, eax
 cmp     rdx, 4
 jae     LBB19_20
 mov     esi, 4
 jmp     LBB19_3
LBB19_18:
 xor     ecx, ecx
 mov     esi, 8
 jmp     LBB19_7
LBB19_19:
 xor     ecx, ecx
[...]

PR #95, opened 2023-12-16
Provided RW impl for fixed arrays.

Frostie314159:
This PR makes reading and writing possible for fixed length arrays of any type implementing `TryFromCtx` or `TryIntoCtx`.

### DIFF
--- a/src/ctx.rs
+++ b/src/ctx.rs
@@ -180,14 +180,14 @@
 //! }
 //! ```
 
-use core::mem::size_of;
+use core::mem::{size_of, zeroed};
 use core::ptr::copy_nonoverlapping;
 use core::{result, str};
 #[cfg(feature = "std")]
 use std::ffi::{CStr, CString};
 
 use crate::endian::Endian;
-use crate::error;
+use crate::{error, Pread, Pwrite};
 
 /// A trait for measuring how large something is; for a byte sequence, it will be its length.
 pub trait MeasureWith<Ctx> {
@@ -778,6 +778,33 @@ impl<'a> TryFromCtx<'a, usize> for &'a [u8] {
         } else {
             Ok((&src[..size], size))
         }
+    }
+}
+
+impl<'a, Ctx: Copy, T: TryFromCtx<'a, Ctx, Error = error::Error>, const N: usize>
+    TryFromCtx<'a, Ctx> for [T; N]
+{
+    type Error = error::Error;
+    fn try_from_ctx(src: &'a [u8], ctx: Ctx) -> Result<(Self, usize), Self::Error> {
+        let mut offset = 0;
+        // SAFETY: We will return an error if filling any of these fails.
+        let mut buf: [T; N] = unsafe { zeroed() };
+
+        for element in buf.iter_mut() {
+            *element = src.gread_with(&mut offset, ctx)?;
+        }
+
+        Ok((buf, offset))
+    }
+}
+impl<Ctx: Copy, T: TryIntoCtx<Ctx, Error = error::Error>, const N: usize> TryIntoCtx<Ctx> for [T; N] {
+    type Error = error::Error;
+    fn try_into_ctx(self, buf: &mut [u8], ctx: Ctx) -> Result<usize, Self::Error> {
+        let mut offset = 0;
+        for element in self {
+            buf.gwrite_with(element, &mut offset, ctx)?;
+        }
+        Ok(offset)
     }
 }
 
--- a/src/ctx.rs
+++ b/src/ctx.rs
@@ -180,7 +180,7 @@
 //! }
 //! ```
 
-use core::mem::{size_of, zeroed};
+use core::mem::{size_of, MaybeUninit};
 use core::ptr::copy_nonoverlapping;
 use core::{result, str};
 #[cfg(feature = "std")]
@@ -787,14 +787,33 @@ impl<'a, Ctx: Copy, T: TryFromCtx<'a, Ctx, Error = error::Error>, const N: usize
     type Error = error::Error;
     fn try_from_ctx(src: &'a [u8], ctx: Ctx) -> Result<(Self, usize), Self::Error> {
         let mut offset = 0;
-        // SAFETY: This will only be read after everything was overwritten.
-        let mut buf: [T; N] = unsafe { zeroed() };
 
-        for element in buf.iter_mut() {
-            *element = src.gread_with(&mut offset, ctx)?;
+        let mut buf: [MaybeUninit<T>; N] = core::array::from_fn(|_| MaybeUninit::uninit());
+
+        let mut error_ctx = None;
+        for (idx, element) in buf.iter_mut().enumerate() {
+            match src.gread_with::<T>(&mut offset, ctx) {
+                Ok(val) => {
+                    *element = MaybeUninit::new(val);
+                }
+                Err(e) => {
+                    error_ctx = Some((e, idx));
+                    break;
+                }
+            }
         }
-
-        Ok((buf, offset))
+        if let Some((e, idx)) = error_ctx {
+            for element in &mut buf[0..idx].iter_mut() {
+                unsafe {
+                    element.assume_init_drop();
+                }
+            }
+            Err(e)
+        } else {
+            // SAFETY: we initialized each element above by preading them out, correctness
+            // of the initialized element is guaranted by pread itself
+            Ok((buf.map(|element| unsafe { element.assume_init() }), offset))
+        }
     }
 }
 impl<Ctx: Copy, T: TryIntoCtx<Ctx, Error = error::Error>, const N: usize> TryIntoCtx<Ctx>
--- a/src/ctx.rs
+++ b/src/ctx.rs
@@ -787,7 +787,7 @@ impl<'a, Ctx: Copy, T: TryFromCtx<'a, Ctx, Error = error::Error>, const N: usize
     type Error = error::Error;
     fn try_from_ctx(src: &'a [u8], ctx: Ctx) -> Result<(Self, usize), Self::Error> {
         let mut offset = 0;
-        // SAFETY: We will return an error if filling any of these fails.
+        // SAFETY: This will only be read after everything was overwritten.
         let mut buf: [T; N] = unsafe { zeroed() };
 
         for element in buf.iter_mut() {
--- a/src/ctx.rs
+++ b/src/ctx.rs
@@ -797,7 +797,9 @@ impl<'a, Ctx: Copy, T: TryFromCtx<'a, Ctx, Error = error::Error>, const N: usize
         Ok((buf, offset))
     }
 }
-impl<Ctx: Copy, T: TryIntoCtx<Ctx, Error = error::Error>, const N: usize> TryIntoCtx<Ctx> for [T; N] {
+impl<Ctx: Copy, T: TryIntoCtx<Ctx, Error = error::Error>, const N: usize> TryIntoCtx<Ctx>
+    for [T; N]
+{
     type Error = error::Error;
     fn try_into_ctx(self, buf: &mut [u8], ctx: Ctx) -> Result<usize, Self::Error> {
         let mut offset = 0;
--- a/src/ctx.rs
+++ b/src/ctx.rs
@@ -804,6 +804,8 @@ impl<'a, Ctx: Copy, T: TryFromCtx<'a, Ctx, Error = error::Error>, const N: usize
         }
         if let Some((e, idx)) = error_ctx {
             for element in &mut buf[0..idx].iter_mut() {
+                // SAFETY: Any element upto idx must have already been initialized, since
+                // we iterate until we encounter an error.
                 unsafe {
                     element.assume_init_drop();
                 }
--- a/tests/api.rs
+++ b/tests/api.rs
@@ -290,3 +290,14 @@ fn cwrite_api_customtype() {
     assert_eq!({ bar.foo }, -1);
     assert_eq!({ bar.bar }, 0xdeadbeef);
 }
+
+#[test]
+fn test_fixed_array_rw() {
+    use scroll::{Pread, Pwrite, Endian, ctx::MeasureWith};
+    let bytes = [0x37, 0x13, 0x37, 0x13];
+    assert_eq!(bytes.measure_with(&()), 4);
+    assert_eq!(bytes.pread_with::<[u16; 2]>(0, Endian::Little).unwrap(), [0x1337u16, 0x1337]);
+    let mut buf = [0x00u8; 4];
+    buf.pwrite([0x1337u16, 0x1337], 0).unwrap();
+    assert_eq!(buf, bytes);
+}
--- a/tests/api.rs
+++ b/tests/api.rs
@@ -284,7 +284,7 @@ impl Drop for VecDrop {
 #[test]
 fn test_fixed_array_str() {
     use scroll::Pread;
-    let bytes = Box::new([0x45, 0x42, 0x0, 0x45, 0x41]);
+    let bytes = [0x45, 0x42, 0x0, 0x45, 0x41];
     let res = bytes.pread_with::<[BytesDrop; 2]>(0, Default::default());
     println!("{res:?}");
     assert!(res.is_ok());
@@ -293,7 +293,7 @@ fn test_fixed_array_str() {
 #[test]
 fn test_fixed_array_string() {
     use scroll::Pread;
-    let bytes = Box::new([0x45, 0x42, 0x0, 0x45, 0x41, 0x0]);
+    let bytes = [0x45, 0x42, 0x0, 0x45, 0x41, 0x0];
     let res = bytes.pread_with::<[VecDrop; 2]>(0, Default::default());
     println!("{res:?}");
     assert!(res.is_ok());
--- a/tests/api.rs
+++ b/tests/api.rs
@@ -226,72 +226,77 @@ impl scroll::ctx::FromCtx<scroll::Endian> for Bar {
         }
     }
 }
+type MyCtx = ();
+
 #[derive(Debug)]
-struct StrDrop<'a> {
+struct BytesDrop<'a> {
     #[allow(unused)]
-    s: &'a str,
+    s: &'a [u8],
 }
 
-impl<'a> scroll::ctx::TryFromCtx<'a, scroll::Endian> for StrDrop<'a> {
+impl<'a> scroll::ctx::TryFromCtx<'a, MyCtx> for BytesDrop<'a> {
     type Error = scroll::Error;
     fn try_from_ctx(
-        bytes: &'a [u8],
-        _ctx: scroll::Endian,
+        src: &'a [u8],
+        _ctx: MyCtx,
     ) -> ::std::result::Result<(Self, usize), Self::Error> {
-        let mut offset = 0;
-        let s = bytes.gread(&mut offset)?;
-        Ok((Self { s }, offset))
+        let len = src.iter().take_while(|c| **c != 0x0).count();
+        if len > src.len() {
+            return Err(scroll::Error::TooBig {
+                size: len,
+                len: src.len(),
+            });
+        };
+        Ok((Self { s: &src[..len] }, len))
     }
 }
 
-impl<'a> Drop for StrDrop<'a> {
+impl<'a> Drop for BytesDrop<'a> {
     fn drop(&mut self) {
         println!("Dropping {self:?}");
     }
 }
 
 #[derive(Debug)]
-struct StringDrop {
+struct VecDrop {
     #[allow(unused)]
-    s: String,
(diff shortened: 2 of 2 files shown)


Frostie314159:
I finally managed to implement this without the whole merge conflict fiasko and as a little bonus made it work for every type. I'm aware, that this uses unsafe, but I think my intuition, that unsafe is necessarily bad was wrong. If we just use it sparingly and with safety comments we should be good to go.

src/ctx.rs
```diff
@@ -781,6 +781,35 @@ impl<'a> TryFromCtx<'a, usize> for &'a [u8] {
...
 }
 
+impl<'a, Ctx: Copy, T: TryFromCtx<'a, Ctx, Error = error::Error>, const N: usize>
+    TryFromCtx<'a, Ctx> for [T; N]
+{
+    type Error = error::Error;
+    fn try_from_ctx(src: &'a [u8], ctx: Ctx) -> Result<(Self, usize), Self::Error> {
+        let mut offset = 0;
+        // SAFETY: This will only be read after everything was overwritten.
+        let mut buf: [T; N] = unsafe { zeroed() };
```
m4b, line 791: this is not actually safe; `*element` line below will call the drop function on `T`, this is a typical pitfall of unsafe rust and using `mem::zeroed`, and hard to catch.

if this were e.g., a String, or some other complex object, it could theoretically segfault during it's drop routine

You have two options:
1. if you want to stick with unsafe, the proper solution is to use `mem::MaybeUninit`, with a write-only initialization of an array being the classic example/usecase.
2. if you don't want to stick with unsafe, i believe you could add a default bound to T, and then do, e.g.: `let mut buf: [T; N] = [T::default(); N]`

I'm ok with MaybeUninit approach, as it is less burdensome on the implementator to require default on their type, however we have to be extra careful during review that it is safe + sound :)

m4b (changes requested):
need to address unsafe use of zeroed as per my comment

src/ctx.rs
Frostie314159: I'm currently working on doing this with `MaybeUninit`, but I don't know how to turn `[MaybeUninit<T>;N]` into `[T;N]` due to the layout guarantees, transmute would be an option.

m4b: Somewhat related, a fully safe impl that requires a SizeWith bound (e.g., no sizes depending on the ctx) which will rule out dynamically sized things like slices would look something like:
```rust
impl<
        'a,
        Ctx: Copy,
        T: TryFromCtx<'a, Ctx, Error = error::Error> + SizeWith<Ctx>,
        const N: usize,
    > TryFromCtx<'a, Ctx> for [T; N]
{
    type Error = error::Error;
    fn try_from_ctx(src: &'a [u8], ctx: Ctx) -> Result<(Self, usize), Self::Error> {
        let nbytes_required = N * T::size_with(&ctx);
        if nbytes_required > src.len() {
            panic!();
        }
        let buf: [T; N] =
            std::array::from_fn(|i| src.pread_with(i * T::size_with(&ctx), ctx).unwrap());

        Ok((buf, nbytes_required))
    }
}
```

I think the use of unsafe and `MaybeUninit` though is actually justified here, since your impl _is_ the "initializer"; again though we must be very careful during review we don't introduce unsoundness in this part of the library

Frostie314159: The issue is, that for one we are not handling errors properly, which will be hard, without using try_from_fn which is unstable and on the other hand can't reasonably introduce a Default bound to all types T.

Frostie314159: I'm currently at an implementation, which looks like this:
```rust
impl<'a, Ctx: Copy, T: TryFromCtx<'a, Ctx, Error = error::Error>, const N: usize>
    TryFromCtx<'a, Ctx> for [T; N]
{
    type Error = error::Error;
    fn try_from_ctx(src: &'a [u8], ctx: Ctx) -> Result<(Self, usize), Self::Error> {
        let mut offset = 0;

        // This is taken from the implementation of `MaybeUninit::uninit_array`, which is currently unstable.
        // SAFETY: An uninitialized `[MaybeUninit<_>; LEN]` is valid.
        let mut buf = unsafe { MaybeUninit::<[MaybeUninit<T>; N]>::uninit().assume_init() };

        for element in buf.iter_mut() {
            *element = MaybeUninit::new(src.gread_with::<T>(&mut offset, ctx)?);
        }

        Ok((buf.map(|element| unsafe { element.assume_init() }), offset))
    }
}
```
The array initialization was taken from `MaybeUninit::uninit_array`, which is unstable. This avoids having to introduce a `Copy` bound on T.

m4b: This seems like the correct approach, although if we did add a copy bound I don’t think we need any unsafe at all.

One thing that does worry me is that if gread fails in that loop, any objects which got put in the array from earlier in the loop before failure, are never dropped. This is maybe ok but it does worry me a bit. There’s a way to fix this but it’s not great (reiterate back up to the failing index and assume_init it and that handles the drop as far as I know)

m4b: Ah, i see what you mean now; yes we still need to `MaybeUninit` the array itself, which requires unsafe, if we have `Copy` bound; but we'd still have unsafe; in order to do full unsafe requires `Default`; i think what you've written is ok, though a `Copy` bound would ensure the type is "safer", but not quite enough.

m4b: Here's another version with only 1 unsafe block, although there's a good chance the array is double-initialized, i.e., first from `from_fn` (likely a memset) then the pread loop, (would need to check assembly):
```rust
impl<'a, Ctx: Copy, T: TryFromCtx<'a, Ctx, Error = error::Error>, const N: usize>
    TryFromCtx<'a, Ctx> for [T; N]
{
    type Error = error::Error;
    fn try_from_ctx(src: &'a [u8], ctx: Ctx) -> Result<(Self, usize), Self::Error> {
        let mut offset = 0;
        use std::mem::MaybeUninit;

        let mut buf: [MaybeUninit<T>; N] = std::array::from_fn(|_| MaybeUninit::uninit());

        for element in buf.iter_mut() {
            *element = MaybeUninit::new(src.gread_with::<T>(&mut offset, ctx)?);
        }

        // SAFETY: we initialized each element above by preading them out, correctness
        // of the initialized element is guaranted by pread itself
        Ok((buf.map(|element| unsafe { element.assume_init() }), offset))
    }
}
```

i've also written a custom type with a custom drop impl and verified drop is only run once for each element (after the array leaves scope) whereas with the zeroed version, it is run twice (the `*element` deref as i mentioned)

Frostie314159: The array init looks pretty good to me, my biggest worry is `Drop` not being called, since `T` could potentially and very realistically contain heap allocated objects like `Box` or `Vec` which wouldn't be freed properly.

m4b: And here is a version that drops the elements if we fail while preading; also for some reason i can't get transmute working, which should be a free cast, according to MaybeUninit docs, i think perhaps it doesn't work with const generics (yet)?

```rust
impl<'a, Ctx: Copy, T: TryFromCtx<'a, Ctx, Error = error::Error>, const N: usize>
    TryFromCtx<'a, Ctx> for [T; N]
{
    type Error = error::Error;
    fn try_from_ctx(src: &'a [u8], ctx: Ctx) -> Result<(Self, usize), Self::Error> {
        let mut offset = 0;
        use std::mem::MaybeUninit;

        let mut buf: [MaybeUninit<T>; N] = std::array::from_fn(|_| MaybeUninit::uninit());

        let mut error_ctx = None;
        for (idx, element) in buf.iter_mut().enumerate() {
            match src.gread_with::<T>(&mut offset, ctx) {
                Ok(val) => {
                    *element = MaybeUninit::new(val);
                }
                Err(e) => {
                    error_ctx = Some((e, idx));
                }
            }
        }
        if let Some((e, idx)) = error_ctx {
            for element in &mut buf[0..idx].iter_mut() {
                unsafe {
                    element.assume_init_drop();
                }
            }
            Err(e)
        } else {
            // SAFETY: we initialized each element above by preading them out, correctness
            // of the initialized element is guaranted by pread itself
            Ok((buf.map(|element| unsafe { element.assume_init() }), offset))
        }
    }
}
```

Frostie314159: I had trouble with `transmute` to, which is why I used `[T; N]::map` instead, but this map will with certainty be optimised out, since the function calls doesn't actually do anything. So far this looks great.

m4b:
I've never seen this error before, looks like transmute doesn't work with const generics, which is a bit sad:
```
error[E0512]: cannot transmute between types of different sizes, or dependently-sized types
   --> src/ctx.rs:815:26
    |
815 |             Ok((unsafe { std::mem::transmute::<_, Self>(buf) }, offset))
    |                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: source type: `[MaybeUninit<T>; N]` (this type does not have a fixed size)
    = note: target type: `[T; N]` (this type does not have a fixed size)
```

src/ctx.rs
Frostie314159: The only thing that is missing, is that the first loop doesn't break on an error.

Frostie314159: I just added that and have this impl now:
```rust
impl<'a, Ctx: Copy, T: TryFromCtx<'a, Ctx, Error = error::Error>, const N: usize>
    TryFromCtx<'a, Ctx> for [T; N]
{
    type Error = error::Error;
    fn try_from_ctx(src: &'a [u8], ctx: Ctx) -> Result<(Self, usize), Self::Error> {
        let mut offset = 0;

        let mut buf: [MaybeUninit<T>; N] = core::array::from_fn(|_| MaybeUninit::uninit());

        let mut error_ctx = None;
        for (idx, element) in buf.iter_mut().enumerate() {
            match src.gread_with::<T>(&mut offset, ctx) {
                Ok(val) => {
                    *element = MaybeUninit::new(val);
                }
                Err(e) => {
                    error_ctx = Some((e, idx));
                    break;
                }
            }
        }
        if let Some((e, idx)) = error_ctx {
            for element in &mut buf[0..idx].iter_mut() {
                unsafe {
                    element.assume_init_drop();
                }
            }
            Err(e)
        } else {
            // SAFETY: we initialized each element above by preading them out, correctness
            // of the initialized element is guaranted by pread itself
            Ok((buf.map(|element| unsafe { element.assume_init() }), offset))
        }

    }
}
```

m4b: yes push this; good catch on that `break` too, that would have been a serious error! :)

Frostie314159: I just pushed it, kudos for spotting the unsoundness.

m4b:
btw i've been testing with this code, helpful to see where drops occur:
```rust
#[derive(Debug)]
struct StrDrop<'a> {
    #[allow(unused)]
    s: &'a str,
}

impl<'a> scroll::ctx::TryFromCtx<'a, scroll::Endian> for StrDrop<'a> {
    type Error = scroll::Error;
    fn try_from_ctx(
        bytes: &'a [u8],
        _ctx: scroll::Endian,
    ) -> ::std::result::Result<(Self, usize), Self::Error> {
        let mut offset = 0;
        let s = bytes.gread(&mut offset)?;
        Ok((Self { s }, offset))
    }
}

impl<'a> Drop for StrDrop<'a> {
    fn drop(&mut self) {
        println!("Dropping {self:?}");
    }
}

#[derive(Debug)]
struct StringDrop {
    #[allow(unused)]
    s: String,
}

impl<'a> scroll::ctx::TryFromCtx<'a, scroll::Endian> for StringDrop {
    type Error = scroll::Error;
    fn try_from_ctx(
        bytes: &'a [u8],
        _ctx: scroll::Endian,
    ) -> ::std::result::Result<(Self, usize), Self::Error> {
        let mut offset = 0;
        let s: &str = bytes.gread(&mut offset)?;
        Ok((Self { s: String::from(s) }, offset))
    }
}

impl Drop for StringDrop {
    fn drop(&mut self) {
        println!("Dropping {self:?}");
        drop(&mut self.s);
    }
}

#[test]
fn test_fixed_array_str() {
    use scroll::Pread;
    let bytes = Box::new([0x45, 0x42, 0x0, 0x45, 0x41]);
    let res = bytes.pread_with::<[StrDrop; 3]>(0, Default::default());
    println!("{res:?}");
    assert!(false);
}

#[test]
fn test_fixed_array_string() {
    use scroll::Pread;
    let bytes = Box::new([0x45, 0x42, 0x0, 0x45, 0x41, 0x0]);
    let res = bytes.pread_with::<[StringDrop; 3]>(0, Default::default());
    println!("{res:?}");
    assert!(false);
}
```

what bothers me is why the `[StringDrop; 3]` is not an erro, but `[StrDrop; 2]` is; need to figure out what's going on there, probably related to the slice logic that changed slightly for pread

Frostie314159:
I also found a [thread](https://users.rust-lang.org/t/transmuting-a-generic-array/45645), which discusses the problem with transmute.

Frostie314159:
I think I found the error, the byte array for `text_fixed_array_str` is missing one zero byte.

tests/api.rs
```diff
@@ -226,6 +226,73 @@ impl scroll::ctx::FromCtx<scroll::Endian> for Bar {
...
+    ) -> ::std::result::Result<(Self, usize), Self::Error> {
+        let mut offset = 0;
+        let s: &str = bytes.gread(&mut offset)?;
+        Ok((Self { s: String::from(s) }, offset))
+    }
+}
+
+impl Drop for StringDrop {
+    fn drop(&mut self) {
+        println!("Dropping {self:?}");
```
m4b, line 273: technically i think you need to call `drop(&mut self.s)`

tests/api.rs
```diff
@@ -226,6 +226,73 @@ impl scroll::ctx::FromCtx<scroll::Endian> for Bar {
...
+    }
+}
+
+#[test]
+#[should_panic]
+fn test_fixed_array_str() {
+    use scroll::Pread;
+    let bytes = [0x45, 0x42, 0x0, 0x45, 0x41];
+    let _res = bytes
+        .pread_with::<[StrDrop; 3]>(0, Default::default())
```
m4b, line 283: i need to understand what's going on here, i'd expect this to panic with an array of 2, since we don't have null terminating byte, and that's the str pread default ctx iirc

tests/api.rs
Frostie314159: I removed it, since the compiler emitted a warning stating, that dropping a reference doesn't actually do anything.

tests/api.rs
Frostie314159: Both of these tests do panic, but since that is the expected behavior, I added `#[should_panic]` to the test.

tests/api.rs
Frostie314159: Here is the relevant doc link: https://doc.rust-lang.org/rustc/lints/listing/warn-by-default.html#dropping-references

tests/api.rs
m4b: But their length is 3.  I’d expect them to panic at 2. Which they do not

Frostie314159: Ah I see what you mean, I'm currently on mobile, but will look into this in an hour or so.

Frostie314159: Ok, I looked into it and ran a test, with this:
```rust
#[test]
fn string_test() {
    use scroll::Pread;
    let bytes = [0x45, 0x42, 0x0, 0x45, 0x41];
    let mut offset = 0;
    bytes.gread_with::<&str>(&mut offset, StrCtx::Delimiter(0)).unwrap();
    bytes.gread_with::<&str>(&mut offset, StrCtx::Delimiter(0)).unwrap();
}
```
This is an issue with the `TryFromCtx` impl on `&str`, it doesn't return an error if no null bytes is encountered, but instead creates a string from as many bytes as are present. This is unrelated to this PR and needs to be fixed separately.

Frostie314159: I opened a PR fixing this #96.

m4b: I don't think we should block this PR on that issue; adding generics is useful enough, and the only thing that's weird is the test; you can just remove the str based tests, though it would be useful to have an allocated, non fixed size/context dependent type (String was for this); I suggest this:

```rust
type MyCtx = ();

#[derive(Debug)]
struct BytesDrop<'a> {
    #[allow(unused)]
    s: &'a [u8],
}

impl<'a> scroll::ctx::TryFromCtx<'a, MyCtx> for BytesDrop<'a> {
    type Error = scroll::Error;
    fn try_from_ctx(
        src: &'a [u8],
        _ctx: MyCtx,
    ) -> ::std::result::Result<(Self, usize), Self::Error> {
        let len = src.iter().take_while(|c| **c != 0x0).count();
        if len > src.len() {
            return Err(scroll::Error::TooBig {
                size: len,
                len: src.len(),
            });
        };
        Ok((Self { s: &src[..len] }, len))
    }
}

impl<'a> Drop for BytesDrop<'a> {
    fn drop(&mut self) {
        println!("Dropping {self:?}");
    }
}

#[derive(Debug)]
struct VecDrop {
    #[allow(unused)]
    s: Vec<u8>,
}

impl<'a> scroll::ctx::TryFromCtx<'a, MyCtx> for VecDrop {
    type Error = scroll::Error;
    fn try_from_ctx(
        bytes: &'a [u8],
        _ctx: MyCtx,
    ) -> ::std::result::Result<(Self, usize), Self::Error> {
        let mut offset = 0;
        let s: BytesDrop = bytes.gread(&mut offset)?;
        Ok((Self { s: Vec::from(s.s) }, offset))
    }
}

impl Drop for VecDrop {
    fn drop(&mut self) {
        println!("Dropping {self:?}");
        drop(&mut self.s);
    }
}

#[test]
fn test_fixed_array_str() {
    use scroll::Pread;
    let bytes = Box::new([0x45, 0x42, 0x0, 0x45, 0x41]);
    let res = bytes.pread_with::<[BytesDrop; 2]>(0, Default::default());
    println!("{res:?}");
    assert!(res.is_ok());
}

#[test]
fn test_fixed_array_string() {
    use scroll::Pread;
    let bytes = Box::new([0x45, 0x42, 0x0, 0x45, 0x41, 0x0]);
    let res = bytes.pread_with::<[VecDrop; 2]>(0, Default::default());
    println!("{res:?}");
    assert!(res.is_ok());
}
```

can add check that the inner bytes of the BytesDrop and VecDrop are as expected, if want. After that, let's merge this!

Frostie314159: I just pushed this. Blocking this will probably not be necessary.

tests/api.rs
```diff
@@ -226,72 +226,77 @@ impl scroll::ctx::FromCtx<scroll::Endian> for Bar {
...
 
 #[test]
-#[should_panic]
 fn test_fixed_array_str() {
     use scroll::Pread;
-    let bytes = [0x45, 0x42, 0x0, 0x45, 0x41];
-    let _res = bytes
-        .pread_with::<[StrDrop; 3]>(0, Default::default())
-        .unwrap();
+    let bytes = Box::new([0x45, 0x42, 0x0, 0x45, 0x41]);
```
m4b, line 287: you didn't need to actually box these bytes but it's fine, doesn't matter

Frostie314159: I actually just copied the tests, but I'm gonna resolve this quickly since it's unnecessary.

m4b (approved):

tests/api.rs
Frostie314159: Fixed with latest commit.

tests/api.rs
m4b: afaik this is a leak in our VecDrop, since you aren't dropping the owned Vec here. The above would apply to the BytesDrop, which is a reference, as you say, which doesn't need to be dropped technically. Since this is a test it doesn't really matter and can be fixed in a followup, but i'm very sure this is a memory leak in this drop impl.

Frostie314159: Actually, the fields of the struct are still dropped by the compiler, see: https://doc.rust-lang.org/std/ops/trait.Drop.html#examples

m4b: ah thank you, I will give it a read, haven't written drop explicitly in a long while, so good to know :)

m4b:
thank you for your patience @Frostie314159 , and thanks for the PR!

Frostie314159:
No problem, the patience was definitely worth it, since we now have a sound implementation.

Frostie314159:
@m4b I just noticed, in one of my projects, that LLVM fails to optimize this properly, when writing byte arrays. We should probably add a warning about that, since it crippled the performance in my case.

m4b:
@Frostie314159 thanks for noting this; there is another issue about llvm failing to optimize before, I wonder if it's related? let me find it; is it worth investigating filing this upstream to rust about perf issue ?

m4b:
Here was the issue, I don't know if it's relevant anymore, it was years ago: https://github.com/m4b/scroll/issues/53

can you paste the assembly if possible, or why you believe the fixed size array is failing to optimize well?

PR #99, opened 2024-06-10
Added SizeWith impl for numeric types without Ctx.

Frostie314159:
This PR introduces `SizeWith` impls for numeric types, without requiring `Endian` as context. It wasn't immediately clear, to me, why the ctx was required.

### DIFF
--- a/src/ctx.rs
+++ b/src/ctx.rs
@@ -750,6 +750,12 @@ macro_rules! sizeof_impl {
                 size_of::<$ty>()
             }
         }
+        impl SizeWith for $ty {
+            #[inline]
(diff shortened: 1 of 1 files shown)


m4b (approved):
I think this should be fine to merge, since the ctx can't affect the size, and we have identical code for Endian

m4b:
@Frostie314159 do you need a release with this and the size with for fixed size arrays? if not necessary I would prefer to wait to see if there's anything else pending, need to re-review the other pending PR here to see if it should be merged

Frostie314159:
It's not urgent, it'd make one section of my code bit neater, but isn't required.

PR #104, opened 2024-08-27
Made `Try{From|Into}Ctx impl` on `CStr` safe.

Frostie314159:
As discussed, I replaced the unsafe `from_bytes_with_nul_unchecked`, with `from_bytes_until_nul` and simplified the writing.

### DIFF
--- a/src/ctx.rs
+++ b/src/ctx.rs
@@ -847,18 +847,11 @@ impl<'a> TryFromCtx<'a> for &'a CStr {
     type Error = error::Error;
     #[inline]
     fn try_from_ctx(src: &'a [u8], _ctx: ()) -> result::Result<(Self, usize), Self::Error> {
-        let null_byte = match src.iter().position(|b| *b == 0) {
-            Some(ix) => ix,
-            None => {
-                return Err(error::Error::BadInput {
-                    size: 0,
-                    msg: "The input doesn't contain a null byte",
-                })
-            }
-        };
-
-        let cstr = unsafe { CStr::from_bytes_with_nul_unchecked(&src[..=null_byte]) };
-        Ok((cstr, null_byte + 1))
+        let cstr = CStr::from_bytes_until_nul(src).map_err(|_| error::Error::BadInput {
+            size: 0,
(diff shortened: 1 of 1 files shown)


Frostie314159:
I think we discussed this previously, but the library feature we're using, requires an MSRV of 1.69.

src/ctx.rs
```diff
@@ -847,18 +847,11 @@ impl<'a> TryFromCtx<'a> for &'a CStr {
...
-            }
-        };
-
-        let cstr = unsafe { CStr::from_bytes_with_nul_unchecked(&src[..=null_byte]) };
-        Ok((cstr, null_byte + 1))
+        let cstr = CStr::from_bytes_until_nul(src).map_err(|_| error::Error::BadInput {
+            size: 0,
+            msg: "The input doesn't contain a null byte",
+        })?;
+        Ok((cstr, cstr.to_bytes_with_nul().len()))
```
m4b, line 854: the docs say:
> Note: This method is currently implemented as a 0-cost cast, but it is planned to alter its definition in the future to perform the length calculation whenever this method is called.

I don't know why they would need to perform the length calculation for this method, ever? So not sure what to think of this, but looking at source it's just an (unsafe, whomp whomp) cast from chars -> u8s, which is fine (and what i'd expect). I wonder if this comment is wrong/outdated.

the only thing I'd say is the previous unsafe version was _slightly_ better since it performed the length calculation (the null byte) while it was also validating, and if for some reason `to_bytes_with_nul` does a linear scan of the string to get the length (as it says it will) that would be a (very?) minor performance regression from the old. But i'm fine with it if it removes direct unsafe and foists it into the stdlib.

m4b (changes requested):
Apologies for the delay, and thanks for adding this; especially the try_into_ctx change is quite nice.

However, the MSRV needs to be updated in order for CI to pass, unfortunately.

Would you mind adding that to this change and then we can merge?

m4b:
@Frostie314159 i think we can safely add this now since we are updating to rust 2024 for the new release

PR #106, opened 2024-10-02
Custom ctx override for derive macro

Easyoakland:
Add attribute to override the context in `scroll-derive` used like the following:
```rust
#[derive(Pread, Pwrite, SizeWith)]
#[repr(C)]
struct Data {
    id: u32,
    timestamp: f64,
    #[scroll(ctx = BE)]
    arr: [u16; 2],
    #[scroll(ctx = CustomCtx::len())]
    custom_ctx: CustomCtx,
}
```
where `ctx` doesn't have to be of type `Endian` and ignores `pread_with`'s (and related) `ctx` argument.

Use case for this includes using the derive macro for structs that don't use `Endian` as their context and deriving many `Network` Endian structs while still being able to use `pread`/`gread`/`pwrite`/`gwrite` without having to specify `Network` Endian every time.

A future extension to this can enable using previous fields in the expression which would enable a pattern like:
```rust
#[derive(Pread, Pwrite, SizeWith)]
#[repr(C)]
struct Data<'b> {
    header: u32,
    #[scroll(ctx = header as usize)]
    body: &'b [u8],
}
```

without having to manually implement the various traits.

Let me know if you're interested in this idea and/or the future extension. Thanks!

### DIFF
--- a/scroll_derive/examples/derive_custom_ctx.rs
+++ b/scroll_derive/examples/derive_custom_ctx.rs
@@ -1,15 +1,22 @@
 use scroll_derive::{Pread, Pwrite, SizeWith};
 
-#[derive(Debug, PartialEq)]
-struct CustomCtx {
-    buf: Vec<u8>,
-}
-impl CustomCtx {
-    fn len() -> usize {
-        3 + 2
+/// An example of using a method as the value for a ctx in a derive.
+struct EndianDependent(Endian);
+impl EndianDependent {
+    fn len(&self) -> usize {
+        match self.0 {
+            scroll::Endian::Little => 5,
+            scroll::Endian::Big => 6,
+        }
     }
 }
-impl<'a> TryFromCtx<'a, usize> for CustomCtx {
+
+#[derive(Debug, PartialEq)]
+struct VariableLengthData {
+    buf: Vec<u8>,
+}
+
+impl<'a> TryFromCtx<'a, usize> for VariableLengthData {
     type Error = scroll::Error;
 
     fn try_from_ctx(from: &'a [u8], ctx: usize) -> Result<(Self, usize), Self::Error> {
@@ -18,7 +25,7 @@ impl<'a> TryFromCtx<'a, usize> for CustomCtx {
         Ok((Self { buf }, *offset))
     }
 }
-impl<'a> TryIntoCtx<usize> for &'a CustomCtx {
+impl<'a> TryIntoCtx<usize> for &'a VariableLengthData {
     type Error = scroll::Error;
     fn try_into_ctx(self, dst: &mut [u8], ctx: usize) -> Result<usize, Self::Error> {
         let offset = &mut 0;
@@ -28,7 +35,7 @@ impl<'a> TryIntoCtx<usize> for &'a CustomCtx {
         Ok(*offset)
     }
 }
-impl SizeWith<usize> for CustomCtx {
+impl SizeWith<usize> for VariableLengthData {
     fn size_with(ctx: &usize) -> usize {
         *ctx
     }
@@ -39,15 +46,20 @@ impl SizeWith<usize> for CustomCtx {
 struct Data {
     id: u32,
     timestamp: f64,
+    // You can fix the ctx regardless of what is passed in.
     #[scroll(ctx = BE)]
     arr: [u16; 2],
-    #[scroll(ctx = CustomCtx::len())]
-    custom_ctx: CustomCtx,
+    // You can use arbitrary expressions for the ctx.
+    // You have access to the `ctx` parameter of the `{pread/gread}_with` inside the expression.
+    // TODO(implement) you have access to previous fields.
+    // TODO(check) will this break structs with fields named `ctx`?.
+    #[scroll(ctx = EndianDependent(ctx.clone()).len())]
+    custom_ctx: VariableLengthData,
 }
 
 use scroll::{
     ctx::{SizeWith, TryFromCtx, TryIntoCtx},
-    Pread, Pwrite, BE, LE,
+    Endian, Pread, Pwrite, BE, LE,
 };
 
 fn main() {
@@ -64,9 +76,6 @@ fn main() {
     let data: Data = bytes.pread_with(0, LE).unwrap();
     let data2: Data = bytes2.pread_with(0, LE).unwrap();
     assert_eq!(data, data2);
-
-    /*
-    let data: Data = bytes.cread_with(0, LE);
-       assert_eq!(data, data2);
-    */
+    // Not enough bytes because of ctx dependant length being too long.
+    assert!(bytes.pread_with::<Data>(0, BE).is_err())
 }
--- a/scroll_derive/examples/derive_custom_ctx.rs
+++ b/scroll_derive/examples/derive_custom_ctx.rs
@@ -1,0 +1,72 @@
+use scroll_derive::{Pread, Pwrite, SizeWith};
+
+#[derive(Debug, PartialEq)]
+struct CustomCtx {
+    buf: Vec<u8>,
+}
+impl CustomCtx {
+    fn len() -> usize {
+        3 + 2
+    }
+}
+impl<'a> TryFromCtx<'a, usize> for CustomCtx {
+    type Error = scroll::Error;
+
+    fn try_from_ctx(from: &'a [u8], ctx: usize) -> Result<(Self, usize), Self::Error> {
+        let offset = &mut 0;
+        let buf = from.gread_with::<&[u8]>(offset, ctx)?.to_owned();
+        Ok((Self { buf }, *offset))
+    }
+}
+impl<'a> TryIntoCtx<usize> for &'a CustomCtx {
+    type Error = scroll::Error;
+    fn try_into_ctx(self, dst: &mut [u8], ctx: usize) -> Result<usize, Self::Error> {
+        let offset = &mut 0;
+        for i in 0..(ctx.min(self.buf.len())) {
+            dst.gwrite(self.buf[i], offset)?;
+        }
+        Ok(*offset)
+    }
+}
+impl SizeWith<usize> for CustomCtx {
+    fn size_with(ctx: &usize) -> usize {
+        *ctx
+    }
+}
+
+#[derive(Debug, PartialEq, Pread, Pwrite, SizeWith)]
+#[repr(C)]
+struct Data {
+    id: u32,
+    timestamp: f64,
+    #[scroll(ctx = BE)]
+    arr: [u16; 2],
+    #[scroll(ctx = CustomCtx::len())]
+    custom_ctx: CustomCtx,
+}
+
+use scroll::{
+    ctx::{SizeWith, TryFromCtx, TryIntoCtx},
+    Pread, Pwrite, BE, LE,
+};
+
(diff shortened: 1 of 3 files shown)


m4b:
omg i wanted something like this for so long, thank you for getting the ball rolling!

scroll_derive/src/lib.rs
```diff
@@ -15,20 +21,63 @@ fn impl_field(ident: &proc_macro2::TokenStream, ty: &syn::Type) -> proc_macro2::
...
+    get_attr("scroll", field).and_then(|x| {
+        // parsed #[scroll..]
+        // `expr` is `None` if the `ctx` key is not used.
+        let mut expr = None;
+        let res = x.parse_nested_meta(|meta| {
+            // parsed #[scroll(..)]
+            if meta.path.is_ident("ctx") {
+                // parsed #[scroll(ctx..)]
+                let value = meta.value()?; // parsed #[scroll(ctx = ..)]
+                expr = Some(value.parse::<syn::Expr>()?); // parsed #[scroll(ctx = expr)]
```
m4b, line 66: is there a reason you don't `return Ok(value.parse::<syn::Expr>()?);` ? this way on the match below you can do `Ok(expr) => expr.into_token_stream()` without the map, which seems slightly more idiomatic to me, but perhaps there's some issue i'm not seeing?

scroll_derive/examples/derive_custom_ctx.rs
```diff
@@ -0,0 +1,72 @@
...
+}
+
+#[derive(Debug, PartialEq, Pread, Pwrite, SizeWith)]
+#[repr(C)]
+struct Data {
+    id: u32,
+    timestamp: f64,
+    #[scroll(ctx = BE)]
+    arr: [u16; 2],
+    #[scroll(ctx = CustomCtx::len())]
```
m4b, line 44: i was surprised you could do arbitrary expressions here; maybe it's the example, but also probably because I haven't written derive code in a while, but:

1. ~I'm a bit confused what's happening here; the ctx is usize? I thought that usize as ctx's were removed in a few scroll's ago, primarily because they give illusion of being safe. so what is going on in this case then, where len() -> usize~ ?
2. not sure the exact usecase  for calling a function like that in the ctx; i'm also wondering out of pure curiosity, could you call a self method on a const or static object to also get an expr that resolves to a ctx?

Re 1., I realized as a I wrote it that the CustomCtx itself is being constructed from a usize ctx, which happens to also be the (static) method on CustomCtx::len that returns a usize.

This is actually a bit confusing and probably not the best example, I would probably change CustomCtx to CustomData, but since you implement SizeWith it is a (custom) Ctx, although it isn't used anywhere _as_ a ctx in parsing later on, so I got confused.

However, these kind of patterns are in PE binaries a lot (a thing gets parsed out using some data, and then that thing is used as the ctx later on, or a prior field is used as the length delimiter, etc.), so maybe it is a useful example? Not sure best way to make it more clear, but it's fine.

Anyway, just kind of rambling at this point :D

m4b (approved):
This is super exciting, thank you for this! It really sets the stage for doing stuff like you mention, e.g., using a previous field as the incoming ctx for next field, etc. But also just being able to swap out field by field a different ctx instead of the derive dominating the whole thing is very useful; it may be possible to use this to go back, for example, to most of goblin's PE struct and now just derive it, since many of them had patterns like this!

The only nit/question I had was about the Option used with `expr` and being idiomatic; if you feel like changing it (and can, and not some oversight by me), feel free; otherwise I'll merge in a few days.

This is *super* awesome, thank you for this!

scroll_derive/examples/derive_custom_ctx.rs
Easyoakland: I'll fix this in a bit. I agree it confusing. I was only trying to demonstrate that arbitrary expressions work, since I was also surprised that's now supported by `syn`. I think I saw somewhere that `serde` used strings when `syn` only supported strings and stays that way for compatibility. The code works by basically copy-pasting the expression entered in the various implementations. It even supports multi-line blocks (if you activate `syn`'s `full` feature) like:
```rs
{
    let a = 3;
    let b = 4;
    a + b
}
```

scroll_derive/src/lib.rs
Easyoakland: `parse_nested_meta` in `syn` has this signature:
```rs
pub fn parse_nested_meta(
        &self,
        logic: impl FnMut(ParseNestedMeta) -> Result<()>,
    ) -> Result<()> {
        self.parse_args_with(meta::parser(logic))
    }
```
So we have to return `Ok(())` on success.

Easyoakland:
I think your comments have been addressed in the last 2 commits. Let me know if not.

Seeing as there's interest I'll elaborate a little.
The idea with using previous fields is instead of the derive making code like:
```rs
let data = Self {
            id: src.gread_with::<u32>(offset, ctx)?,
            timestamp: src.gread_with::<f64>(offset, ctx)?,
            custom_ctx: src.gread_with::<CustomCtx>(offset, USER_SPECIFIED_EXPR)?,
        };
```
it makes code like:
```rs
let id = src.gread_with::<u32>(offset, ctx)?;
let timestamp = src.gread_with::<f64>(offset, ctx)?;
let custom = src.gread_with::<CustomCtx>(offset, USER_SPECIFIED_EXPR)?
let data = Self {
            id,
            timestamp,
            custom,
        };
```
and so `USER_SPECIFIED_EXPR` can use the `id` and `timestamp` and `ctx` identifiers.

The only issue that I see with that approach is it breaks structs which use the derive and have a field named `ctx`.

scroll_derive/examples/derive_custom_ctx.rs
```diff
@@ -0,0 +1,81 @@
...
+#[derive(Debug, PartialEq, Pread, Pwrite, SizeWith)]
+#[repr(C)]
+struct Data {
+    id: u32,
+    timestamp: f64,
+    // You can fix the ctx regardless of what is passed in.
+    #[scroll(ctx = BE)]
+    arr: [u16; 2],
+    // You can use arbitrary expressions for the ctx.
+    // You have access to the `ctx` parameter of the `{pread/gread}_with` inside the expression.
```
m4b, line 53: this is so wild hah, i have to think about usecases/what it means, or if there's some kind of hygiene issue that a different design might not have, but it's very interesting! is the fact that `ctx` is available within the scope of the pread something you have made available or is it a property of derive with field overrides in macros?

scroll_derive/examples/derive_custom_ctx.rs
```diff
@@ -0,0 +1,81 @@
...
+    // TODO(implement) you have access to previous fields.
+    // TODO(check) will this break structs with fields named `ctx`?.
+    #[scroll(ctx = EndianDependent(ctx.clone()).len())]
```
m4b, line 56: one last thing I'll say: i think the example is actually better if EndianDependent is the ctx for VariableLength data, and inside of it's impl, it calls `len()` to get the read/write size, instead of the ctx == read/write size and relying on `len()` being called/passed in here.

That is clearer and more idiomatic scroll, imho, if there is such a thing :D

However, i understand that you are showing off arbitrary expressions are callable in the ctx here, which is very cool and probably worth retaining as you've written it.

Side note: I'm surprised you have to `ctx.clone()` ? `Endian` is Copy and I'd expect to not need that clone there? I also recall that effectively Ctx's generally need to be copy to be usable within pread/gread but maybe I forgot and that bound was relaxed?

Side Side note: if the ctx does need to be cloned, and you don't clone, how terrible is the error message presented to the user (at derive time)? I imagine it could be pretty bad?

m4b (approved):
this is so awesome, thank you for this! Re the other design questions, expanding on usecases for this, e.g., a previous field used as ctx, etc.,, it might be good to first flesh out some possible approaches (assuming there is more than one) and the possible tradeoffs of choosing them, since backwards compat will be something to think about since scroll is effectively "done" (i need to get around to releasing 1.0 :) )

scroll_derive/examples/derive_custom_ctx.rs
Easyoakland: It's because the user's expression is effectively copy-pasted without hygiene and so the parameters of the trait method are accessible. It would take more work to remove this ability, since we'd have to parse for an instance of a `ctx` ident and actively block it.

Easyoakland:
> there's two interesting things happening here; a hardcoded offset is used for the first instance (although I think this is actually wrong, there's an issue open in goblin recently about this); and the second instance uses it's pread offset based on the previous field/result.

I think this would be solved with an `offset` attribute which works like the `ctx` one here. The `offset` identifier would then be available since the derive starts with `let offset = ...`


> There's also the classic cases of (count, items) delimited pairs, e.g., many formats have some fixed size number, usually a u32, which is used to parse out the number of successive elements. Generally the count is discarded in this case, e.g.:

You _could_ make the entire expression to parse a single struct member customizable, but I don't know if that's going to be more readable then implementing the trait manually.

> Lastly, this made me wonder if we can have better derive error messages, e.g., if we fail to parse a field, we could in theory emit something like PE's error message, with the field name known statically, with something like "could not parse " which should in theory be a `&'static str`, maybe we already do something like this though, as I said I haven't thought about/touched the derive code in a while :)

I don't think it currently works that way. It could, though I'm not sure if that would also cause a noticeable binary size increase.

scroll_derive/examples/derive_custom_ctx.rs
Easyoakland:
```
error[E0308]: mismatched types
  --> examples\derive_custom_ctx.rs:56:36
   |
56 |     #[scroll(ctx = EndianDependent(ctx).len())]
   |                    --------------- ^^^ expected `Endian`, found `&Endian`
   |                    |
   |                    arguments to this struct are incorrect
   |
note: tuple struct defined here
  --> examples\derive_custom_ctx.rs:4:8
   |
4  | struct EndianDependent(Endian);
   |        ^^^^^^^^^^^^^^^
help: consider dereferencing the borrow
   |
56 |     #[scroll(ctx = EndianDependent(*ctx).len())]
```
But if you follow the directions:
```
error[E0614]: type `Endian` cannot be dereferenced
  --> examples\derive_custom_ctx.rs:56:36
   |
56 |     #[scroll(ctx = EndianDependent(*ctx).len())]
   |                                    ^^^^
```
The `clone` works because `Endian.clone()` and `(&Endian).clone` both become `Endian`.
The issue is that the `SizeWith` derive makes this code:
```rs
fn size_with(ctx: &::scroll::Endian)
``` 
but the try_{from,into}_ctx derives make this code:
```rs
fn try_into_ctx(
        self,
        dst: &mut [u8],
        ctx: ::scroll::Endian,
    ) ->
```
and
```rs
fn try_from_ctx(
        src: &'a [u8],
        ctx: ::scroll::Endian,
    ) ->
```

Yeah, its a little hacky :)

If `size_with` was by value then `EndianDependant(ctx)` would work.

scroll_derive/examples/derive_custom_ctx.rs
Easyoakland: This also means that if only `Pwrite` derive is used, they won't get an error using `self` or `dst` idents.

m4b:
@Easyoakland released in 0.12.1

PR #114, opened 2025-05-03
Implement core::error::Error

### LINKED ISSUE #111: core::error::Error
`scroll::Error` currently implements [`std::error::Error`](https://doc.rust-lang.org/std/error/trait.Error.html) if the `std` feature is enabled.

This trait is now also available as [`core::error::Error`](https://doc.rust-lang.org/core/error/trait.Error.html) (since Rust 1.81), and it'd be great for some embedded code I'm writing if that could be available. This would mean bumping the Rust version listed in the Cargo manifest though, which I don't know how you feel about.

liclac:
Now that `core::error::Error` is available in stable Rust (as of 1.81.0), we can close #111.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -15,7 +15,7 @@ jobs:
         include:
           - build: MSRV # Minimum supported Rust version, ensure it matches the rust-version in Cargo.toml
             os: ubuntu-latest
-            rust: 1.63
+            rust: 1.81
           - build: stable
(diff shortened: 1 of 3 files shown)


m4b (approved):
this is awesome, thanks! I wonder if it might be a good opportunity to take a critical look at error messages in scroll, they're quite terse and people have complained in the past, i wonder if we can do better?

